PR #59, opened 2015-08-12
user-format change

rssanborn:
update format for boolean items in the post user request table.


rssanborn:
ok, I'll send files. no bloating my stats ;)

Issue #76, opened 2015-08-19
markdown is a mess on REST API Basics page


howarddierking:
http://preview.developer.concur.com/docs/guides/REST-API-Basics.html

Many problems with the markup here

charlieconcur:
Revised markup which may fix several of the issues can be found here: https://github.com/concur/receipt-service/wiki/REST-API-Basics

henryyamamoto-concur:
The same issue appears to be impacting this page as well: 
http://preview.developer.concur.com/docs/guides/index.html

I've checked the markdown and it has the right ``` marks in the right places, but for some reason it's not rendering properly on the site. I've even tried cutting and pasting the code from Charlie's receipt page and no bueno.

Is the CSS being applied based on the front matter "style: reference" or is it only applying it to the api-reference folder?

nicklathe:
After a little research, it looks like Jekyll prefers ~~~ for code blocks instead of ```.

charlieconcur:
Well...the problem with the statement about Jekyll:

* The markdown for http://preview.developer.concur.com/docs/guides/REST-API-Basics.html doesn't use any ``` -- It only uses tabs.
* ~~~ is not a part of the Markdown spec according to http://daringfireball.net/projects/markdown/syntax or https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

I would suggest we make sure our markdown can move seamlessly between GitHub and being run through http://mixu.net/markdown-styles/ (with the --markdown template) without any authoring changes.

nicklathe:
Ok, I'm cool with finding a better way. Currently REST-API-Basics.html does use ```. Here is the raw code: https://raw.githubusercontent.com/concur/developer.concur.com/preview/docs/guides/REST-API-Basics.markdown

nicklathe:
Also, we are using Kramdown, not vanilla Markdown. Kramdown specifies the use of either a 4 space indent or ~~~ for code blocks.

charlieconcur:
The original markdown for this document did not include ``` -- see https://raw.githubusercontent.com/retrosight/common/master/test.md -- so either a tool added the ``` or a human did so. Anybody know for certain it was a human?

Why did we pick Kramdown over vanilla Markdown?

charlieconcur:
[...]

PR #80, opened 2015-08-19
Changed back-ticks to tildes, as Jekyll prefers the later

nicklathe:


charlieconcur:
Yeah, I'm going to strongly disagree with this approach, my friend. ;-) See my comments on #76.

howarddierking:
:shipit: 

looks good - thanks for all the investigation on this

PR #116, opened 2015-08-20
Placeholder text added to forms.

nicklathe:


manage-apps/register.html
```diff
@@ -80,7 +80,7 @@
           <section class="col col-6">
             <label class="label">Website</label>
             <label class="input">
-              <input type="text" name="website" id="website">
+              <input type="text" name="website" id="website" placeholder="https://example.com">
```
howarddierking: should we have the example URL be http rather than https? seems like this would be the more common case.

manage-apps/register.html
```diff
@@ -191,6 +191,7 @@
         equalTo: 'Password and Verify Password must be equal'
       },
       country: { required: 'Please enter your country' },
+      website: { required: 'Please enter a website, with full url including https://' },
```
howarddierking: ditto on scheme

manage-apps/register.html
nicklathe: Ok. Will change to http in both places.

howarddierking:
:shipit:

PR #196, opened 2015-09-02
Update register.html

cskelley:
Clarified the "Before You Start" caution around selecting an email address for the sandbox ID and other copy.


henryyamamoto-concur:
Reviewed the editorial changes and it is good to go.

henryyamamoto-concur:
Fixed #196

PR #198, opened 2015-09-03
Corrected mailto: address for TripLink/travel suppliers

cskelley:
The business development email address for travel and TripLink suppliers is connect@concur.com.


henryyamamoto-concur:
@cskelley May also want to check with the main Concur site if there are uses of the other email address. The one on the site is the one that is currently in the partner sections of the live site.

cskelley:
Both the main site (https://www.concur.com/en-us/connect-platform/suppliers) and the current developer.concur.com site (https://developer.concur.com/get-started/open-booking) use the mailto:connect@concur.com address.

henryyamamoto-concur:
Got it. I was looking here: https://developer.concur.com/go-market/partner-center and the right-hand box on these pages: https://developer.concur.com/get-started/enterprise-apps

PR #204, opened 2015-09-10
Added post function to redirect and login to sandbox; Changed form su…

nicklathe:
…ccess message depending url

### DIFF
--- a/_includes/form-success.html
+++ b/_includes/form-success.html
@@ -1,5 +1,9 @@
 <!-- <div id="form-success" class="form-success-hidden"> -->
 <div id="form-success" class="form-success-hidden alert alert-success alert-dismissible" role="alert">
   <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-  <h3>Thank you. Your request was successfully submitted</h3>
+  {% if page.url == "/manage-apps/register.html" %}
+    {{ <h3>Please wait. Logging into sandbox.</h3> }}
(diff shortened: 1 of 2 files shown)


nicklathe:
@howarddierking 
- I used a setTimeout so the user has time to read the message explaining that they will be logged into sandbox. Similar to how the current site does it.
- I have console.log statements to see if I'm getting the expected data. I'll remove them before production.

_includes/form-success.html
```diff
@@ -1,5 +1,9 @@
 <!-- <div id="form-success" class="form-success-hidden"> -->
 <div id="form-success" class="form-success-hidden alert alert-success alert-dismissible" role="alert">
   <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-  <h3>Thank you. Your request was successfully submitted</h3>
+  {% if page.url == "/manage-apps/register.html" %}
```
howarddierking: basing this decision on the page URL seems slightly brittle in the sense that if we changed the page name, this would break. What if we added a variable in the front matter (something like `formType: sandboxRegistration`) and then check for that in the comparison?

howarddierking:
finished pass

_includes/form-success.html
nicklathe: Now that you mention it, title: Sandbox Registration already exists. I guess I could use that.

howarddierking:
that works

:shipit:

PR #221, opened 2015-10-02
Test post

nicklathe:


nicklathe:
@howarddierking Maybe reaching for straws here, but I slimmed down the hidden form to only what we need, and removed the SetTimeout function. Thought I'd check with you before I merge it in.

PR #232, opened 2015-10-07
Add sitemap.xml generator

howarddierking:
* added sitemap generator from https://github.com/kinnetica/jekyll-plugins
* added basic config for the generator


nicklathe:
:shipit:

PR #244, opened 2015-10-12
Fix GET notifications docs

henryyamamoto-concur:


henryyamamoto-concur:
Need to get API docs for expense notification GET fixed.

henryyamamoto-concur:
Added fix for broken link for Travel Request.

PR #254, opened 2015-10-19
Added asterisk to required fields

nicklathe:
@howarddierking I tried adding a color rule, but ultimately felt the asterisk is more clear.


howarddierking:
works for me

:shipit:

PR #258, opened 2015-10-20
Update booking-resource.markdown

dmclaughlin-concur:
Removed ItinSourceName as a required field and in request as this is set by Itin not the caller.


henryyamamoto-concur:
Looks good. Thanks for updating!

dmclaughlin-concur:
Thanks, looks good on preview...please merge to livesite when able

PR #297, opened 2015-11-19
Update breadcrumb Home link to use site.baseurl

jeffhandley:
It was hard-coded to use / instead of respecting the base url.  The trailing slash is appended within the href.


henryyamamoto-concur:
@nicklathe Can you review and merge? Thanks.

nicklathe:
Looks good. Merging.

PR #304, opened 2015-12-04
Added functionality to receive sandbox messages from backend, and dis…

nicklathe:
…play to user.

@howarddierking still need to test this locally, but wanted to make what I have so far available for review.


nicklathe:
@howarddierking Front-end now receives and displays error message sent from sandbox creation, and stops second call that redirects to sandbox.

manage-apps/register.html
```diff
@@ -204,9 +206,15 @@
           $('#sandboxRegistration button[type="submit"]').attr('disabled', true);
         },
         success: function(data) {
-          console.log(data);
-          $("#sandboxRegistration").addClass('submited');
-          $('#form-success').removeClass('form-success-hidden');
+          if (data.sandboxErr !== 'No Error') {
```
howarddierking: this part seems a little "magic stringy" to me - is there a more standard way to identify errors?

howarddierking:
done with review pass

manage-apps/register.html
nicklathe: Now that I think about it, on the back end, I'm changing null to "No Error." I guess I could just keep it null, and check for that. What do you think?

howarddierking: I was hoping for something more like an HTTP status code

nicklathe: The problem is, the HTTP status code that comes back from the sandbox call is a 200 OK, even though there is an error in the body. I could manufacture a status code I guess, and pass that back?

howarddierking: > I could manufacture a status code I guess, and pass that back

Precisely. One of the value adds that our back-end can provide is a proper HTTP interface, even in the event that CTE does not. This way, we're 1) insulating our Web client from future changes, and 2) not spreading CTE's poor API design.

nicklathe: Cool, makes perfect sense. Thanks!

nicklathe:
@howarddierking Updated code to receive status code from the backend on errors now.

howarddierking:
:shipit:

henryyamamoto-concur:
I've been testing preview vs. livesite and I'm getting consistent success on preview.

We may need to add more copy at the bottom for troubleshooting tips:
* Email must be unique and cannot be one used previously.
* Password must be seven characters or longer
* Be sure you are logged out of Concur or Concur Sandbox.

nicklathe:
Merging. Had to pull down changes in preview first, so it could be auto merged.

PR #317, opened 2015-12-25
Add description about implementation env

TomoTsukui:


henryyamamoto-concur:
Concur Implementation has been getting away from using test sites for over 5 years.  They charge clients for the use of a test site.
Using test entities is also a burden on the partner:
* They have to juggle another app key and secret since test entities can only use a private app
* They have to adjust their entire app to point to the implementation.concursolutions.com url

TomoTsukui:
thanks, i understood.

PR #336, opened 2016-01-20
toggle disable on state dropdown field

merittrompeter:
Disables the 'state' select field on the newsletter signup form when the US is not selected.


nicklathe:
Looks good to me. :shipit:. Wait on @howarddierking 's final word.

PR #343, opened 2016-01-25
minimize side scrolling and change version folder names

merittrompeter:
Minimized side scrolling on swagger tables by making the input and textarea fields responsive; changed folder names from format 'V30' to 'V3-0'


assets/css/style.css
```diff
@@ -2234,3 +2234,12 @@ img.img-mouse {
...
 }
+
+/*Swagger Tables - Prevent Side Scrolling*/
+.parameter{
+  width: 100% !important;
+}
+
+.body-textarea {
+  width: 100% !important;
+}
```
nicklathe: I'm very hesitant to use !important. I did a little digging into the swagger css. In swagger-ui/css/screen.css on line 462, what if we just change width from 300px to 200px? This would make our table slimmer, and get rid of the side scrolling issue.

nicklathe: Well, now that I play with it, that only affects GET table bodies. Let's go with !important for now. But, can we move this into concur_assets/css/swagger-ui-styles.css, to keep things clean? Seems like it would be better suited in this css file.

merittrompeter: I agree, I was worried about important too - and I will definitely move it!

nicklathe:
I say :shipit:.

merittrompeter:
Great, thanks! And I'll continue to try and find a way that does not use !important

FYI these two additional commits update additional v30s to v3-0s; fix a typo; and update the href of the logo at the bottom of the page.

nicklathe:
Additional changes look good. :shipit:!

PR #349, opened 2016-01-28
Update expense-entry.markdown

dstaab:


api-reference/expense/expense-report/expense-entry.markdown
```diff
@@ -186,5 +186,5 @@ Name | Type | Format | Description
 `UnitOfMeasure`	|	`string`	|	-	|	**Required** The unit of measure for distance and odometer values. Possible values: M - miles, K - kilometers
 
 
-
+NOTE: Clients that have Car Configurations that include variable rates are not currently supported.  We only support Car Configurations that include Personal Car One-Rate definitions where Google Maps in not set as mandatory.
```
howarddierking: > where Google Maps in not set as mandatory

"in" == "is"?

dstaab: good catch...that should be "is"

henryyamamoto-concur: Go ahead and make the change and we'll merge it. Thanks!

PR #352, opened 2016-01-28
developer blog with archive and tags

merittrompeter:
Here's some basic blog functionality that you can find at /blog:
- new posts can be created by creating .markdown file in blog/_posts
- tags enabled though YAML in .markdown file
- blog index has pagination (5 excerpts per page)
- most recent 10 posts in left nav with tags (also linked)
- tags index and archives index

Input about formatting, what features are needed, how it should be laid out, etc. would be much appreciated!


nicklathe:
This looks really good! I've tried it locally, and really like it. Now we just need some content!

nicklathe:
@howarddierking do you want to take a look at this? Should we get it in preview for now?

henryyamamoto-concur:
If we can get the two blog posts in, we should get this to preview. I'll right a post on how to update pages later.

merittrompeter:
Those posts are in, so I think we're ready to go - should I go ahead and merge?

henryyamamoto-concur:
We'll give @howarddierking another day to take a look or say he's good if Nick looked at it.

Gemfile
```diff
@@ -1,3 +1,7 @@
 source 'https://rubygems.org'
 gem 'github-pages'
+<<<<<<< HEAD
```
howarddierking: These merge conflict markers need to be resolved - otherwise, the site won't build

Gemfile.lock
```diff
@@ -51,8 +58,8 @@ GEM
     html-pipeline (2.3.0)
       activesupport (>= 2, < 5)
       nokogiri (>= 1.4)
-    i18n (0.7.0)
```
howarddierking: when trying to build locally, getting the error `Your Gemfile.lock is corrupt. The following gem is missing from the DEPENDENCIES section: 'i18n'` - guessing it may have something to do with this. I'm going to try blowing away and recreating the lock file so that everything re-resolves.

howarddierking:
I'm currently unable to build locally - getting the following error: 

`jekyll 3.0.2 | Error:  undefined method `name' for #<Jekyll::Document:0x007ff6fa2039f8>`

Not sure what version of Jekyll everyone else is using - that could be part of the issue. From looking at this PR, the most likely candidate is the tags feature.

Gemfile
```diff
@@ -1,3 +1,7 @@
...
+gem 'jekyll-paginate'
```
howarddierking: I don't think that you need to add this directly as it's pulled in by the github-pages gem

merittrompeter:
Thank you! I'm not having trouble locally but I'm running an older version of Jekyll; I'll try updating and testing it again.

nicklathe:
:shipit:

PR #353, opened 2016-01-28
separate acconect utility and web test tool markdown files

merittrompeter:
Updated the markdown to remove repetition in tools section, renamed the markdown titles to preserve using the left-nav as table of contents.

Is this an okay alternative to one file? (It seemed a little more user friendly to me this way!)


nicklathe:
:shipit:

PR #361, opened 2016-02-03
Added reference links

carolrussell:
Reference section:  added a link below each reference "topic" so readers can return to the Reference topics list.
Reference Itinerary section:  added links to each itinerary element since the list is so long


carolrussell:
Please process changes.  Thanks.

PR #362, opened 2016-02-03
Enable Swagger test functionality cross-origin

howarddierking:
* removed CORS header generation from code as this moves to the edge
* forms and script for calling APIs from SwaggerUI
* Vagrant setup for running locally


nicklathe:
Pulled this down and ran it locally. Works like a charm. :shipit:

henryyamamoto-concur:
I think this pull request can be closed because it was superseded by another one, but wanted to check first.

PR #372, opened 2016-02-10
Update partner-applications.markdown

dstaab:


henryyamamoto-concur:
@dstaab There appears to be a type in your second block of changes the ":" is not appearing as it should. It shows up as "...". Can you check the markdown and make those changes?

henryyamamoto-concur:
Also, I'm noticing you're changing the final approved version from Cadence. Does the copy not flow well now that you see it laid out on the site? Or are these additional catches?

dstaab:
these changes pertain to a doc that was already existing (in ref to your email requesting us to provide feedback on the "Guides" page.  It is difficult to pinpoint where you need me to look ("2nd block of changes")

PR #374, opened 2016-02-11
navbar bottom

merittrompeter:
updated footer so that it does not pull up when there is minimal content on the page; also wrapped the blog layout in the proper classes so that the update would work on the blog


henryyamamoto-concur:
@howarddierking, @ncronquist Can we get a review to ship it?

henryyamamoto-concur:
@howarddierking or @nicklathe Ping, ping.

_includes/footer.html
```diff
@@ -35,9 +35,14 @@
...
                             <p class="md-margin-bottom-40">
-                              <a href="https://www.concur.com/blog/" target="_blank" class="blog-link">
+                              <a href="{{ site.baseurl }}/blog/" class="blog-link">
                                 Read the latest news for apis, features, and best practices of the Concur Platform.
                               </a>
+                              <br>
+                              <a href="{{ site.posts[0].url }}">{{ site.posts[0].title}}</a>
+                              <br>
+                              <a href="{{ site.posts[1].url }}">{{ site.posts[1].title}}</a>
+                              <a></a>
```
nicklathe: I think we should removed the first a tag ("Read the latest news for apis..."), and just show the 3 most recent blogs (2 for now since that's all we have).

Also, add the blog-link class to the other a tags so color is white instead of blue.

concur_assets/css/style.css
```diff
@@ -111,3 +111,39 @@ ul.blog-tags a:hover {
 .col-md-9 li a{
   color: #0275BC;
 }
+
+/* Navbar Pull Bottom */
```
nicklathe: Footer looks great. Just to clarify, this comment should say Footer Pull Bottom, right?

_includes/footer.html
merittrompeter: Good idea, that looks much better!

concur_assets/css/style.css
merittrompeter: Absolutely! My bad!

nicklathe:
I made some styling changes so the Blog list in the footer matches the section to it's left, About Concur. I ran the changes by @merittrompeter. This PR gets a :shipit: from me.

PR #389, opened 2016-02-18
 TVR-5290 modify the link for Workflow Action and enter space to render good form tittles

benedictadesouza:
Fix a bug related to link navigation for Workflow action, and fix pages titles which were not rendered correctly.


henryyamamoto-concur:
Thanks for these fixes @benedictadesouza

PR #390, opened 2016-02-18
fix API nav

merittrompeter:
changed href to fix API dropdown so it works on mobile


henryyamamoto-concur:
@howarddierking or @nicklathe Can we get a check to so we can merge this fix? Thanks.

howarddierking:
sorry for the delay - :shipit:

PR #394, opened 2016-02-20
Vagrantfile improvements and enabled test proxy in a limited capacity

howarddierking:
- switched Vagrantfile to use the virtualbox provider (since it's free and this is an OSS repo)
- simplified Vagrant host to use NGINX directly on the guest rather than in Docker
- removed the hidden field for API key on API explorer pages
- updated the expense group configurations swagger document to specify https://developer.concur.com/api-explorer-proxy/* as the URL base for testing
- NGINX configuration to proxy swagger test calls to www.concursolutions.com


bhague1281:
Tested the Vagrantfile improvements and confirmed working locally.

* Changed Swagger to match existing files.
* Removed reference changes, as the API key field was already visible.

PR #404, opened 2016-02-24
display message on timeout

merittrompeter:
The empty error in issue #56 seems to be the result of a timeout in the auto login (issue #62): it appears when the account has been created, but the auto login times out.

Since we're planning to change this process soon anyway, this fix simply catches the error on the front and directs the user to proceed to concursolutions.com and login with the credentials they just created. 

We can also look into changing the the timeout settings on AWS. Thoughts?

### DIFF
--- a/manage-apps/register.html
+++ b/manage-apps/register.html
@@ -206,13 +206,12 @@ title: Sandbox Registration
           $('#sandboxRegistration button[type="submit"]').attr('disabled', true);
         },
         error: function(res) {
-          console.log('******',res)
-          if (res.statusText === 'error') {
-            $('#form-error > h3').append('Login Timeout Error: Please proceed to <a href="https://www.concursolutions.com">www.concursolutions.com</a> and login with the credentials submitted below.');
-          }
           $('#status-modal').modal('hide');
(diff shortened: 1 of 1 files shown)


manage-apps/register.html
```diff
@@ -206,6 +206,10 @@
           $('#sandboxRegistration button[type="submit"]').attr('disabled', true);
         },
         error: function(res) {
+          console.log('******',res)
+          if (res.statusText === 'error') {
```
nicklathe: I wonder if there is ever a case where we would end up showing both this error message and the one on line 215? I'm not sure what the error object looks like in a login timeout case.

merittrompeter: You're very right. I didn't encounter any double errors but that certainly doesn't mean they couldn't happen! I've committed a tweak that should help, but I haven't been able to get an error yet in my local test to make sure - I'll let you know when I do.

merittrompeter: Test went through and it looks like this change should address potential double error messages.

nicklathe:
Cool. :shipit:

PR #411, opened 2016-03-01
update Marketing section of Partner Resources page

merittrompeter:
Update copy and change GA tracking link to format requested.

(Just want to make sure this is correct with the one  form and the one GA event.)


henryyamamoto-concur:
Under the "App Center Marketing Kit" section, we need to add the download button with PDF icon for this doc that's attached to the issue: 2016_02_02_Concur-partner-toolkit_R5.pdf

merittrompeter:
Right! Just like it says, apologies - update made.

PR #415, opened 2016-03-02
change tracking names to file names

merittrompeter:
updating GA tracking to desired format


henryyamamoto-concur:
I think this is pretty safe, but I'm not a doctor. @nicklathe

nicklathe:
Looks good. It looks like this is just getting more granular on what exactly was clicked/viewed/whatever?

merittrompeter:
Absolutely, instead of the generic titles I had in before, we're substituting the file names of the files.

nicklathe:
Cool. :shipit:

PR #422, opened 2016-03-07
updating YAML US value to USA

merittrompeter:
Marketing would like the value of the United States country choice to be USA instead of US.


nicklathe:
Looks good to me. :shipit:

PR #426, opened 2016-03-10
mask Swagger Proxy

merittrompeter:
Replace outgoing call url; edit response when loaded in UI

### DIFF
--- a/swagger-ui/swagger-ui.js
+++ b/swagger-ui/swagger-ui.js
@@ -31813,7 +31813,6 @@ SwaggerUi.Views.OperationView = Backbone.View.extend({
   },
 
   // puts the response data in UI
-
   showStatus: function(response) {
     var url, content;
     if (response.content === undefined) {
@@ -31926,6 +31925,7 @@ SwaggerUi.Views.OperationView = Backbone.View.extend({
     $('.response', $(this.el)).slideDown();
     $('.response_hider', $(this.el)).show();
     $('.response_throbber', $(this.el)).hide();
+
 
     //adds curl output
     var curlCommand = this.model.asCurl(this.map);
--- a/swagger-ui/swagger-ui.js
+++ b/swagger-ui/swagger-ui.js
@@ -4,7 +4,6 @@
  * @link http://swagger.io
  * @license Apache-2.0
  */
-
 (function(){this["Handlebars"] = this["Handlebars"] || {};
 this["Handlebars"]["templates"] = this["Handlebars"]["templates"] || {};
 this["Handlebars"]["templates"]["apikey_button_view"] = Handlebars.template({"compiler":[6,">= 2.0.0-beta.1"],"main":function(depth0,helpers,partials,data) {
@@ -42,6 +41,7 @@ this["Handlebars"]["templates"]["content_type"] = Handlebars.template({"1":funct
   return buffer + "</select>\n";
 },"useData":true});
 'use strict';
+
 
 $(function() {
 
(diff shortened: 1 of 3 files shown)


swagger-ui/swagger-ui.js
```diff
@@ -30693,7 +30705,7 @@ Emitter.prototype.hasListeners = function(event){
  * TODO: combatible error handling?
  */
 
-module.exports = function(arr, fn, initial){  
+module.exports = function(arr, fn, initial){
```
howarddierking: if this only modifies spacing, I wouldn't include it as this isn't a file that we maintain.

swagger-ui/swagger-ui.js
```diff
@@ -42,6 +42,18 @@ this["Handlebars"]["templates"]["content_type"] = Handlebars.template({"1":funct
 },"useData":true});
 'use strict';
 
+var replaceResponse = function(a, b, element) {
+  if (!element) element = document.body;
+  var nodes = element.childNodes;
+  for (var n=0; n<(nodes.length ? nodes.length : 0); n++) {
+    if (nodes[n].nodeType == Node.TEXT_NODE) {
+      var reg = new RegExp(a, 'gi');
```
howarddierking: If using a regex, declare it in an outer scope since building a regex is relatively expensive

swagger-ui/swagger-ui.js
```diff
@@ -42,6 +42,18 @@ this["Handlebars"]["templates"]["content_type"] = Handlebars.template({"1":funct
...
+      nodes[n].textContent = nodes[n].textContent.replace(reg, b);
+    } else {
+      replaceResponse(a, b, nodes[n]);
```
howarddierking: hmm - could we use jquery for finding the node that matches the criteria (e.g. https://api.jquery.com/contains-selector/)? Just wondering whether there's a more performant approach than recursive decent on the DOM.

swagger-ui/swagger-ui.js
```diff
@@ -42,6 +42,18 @@ this["Handlebars"]["templates"]["content_type"] = Handlebars.template({"1":funct
 },"useData":true});
 'use strict';
 
+var replaceResponse = function(a, b, element) {
```
howarddierking: consider putting this function in its own Javascript document and then modify this file only to call it

swagger-ui/swagger-ui.js
nicklathe: Yeah, jQuery is great for this. Something like this: http://jsfiddle.net/jamesallardice/wQns7/
I think this should work for what we are trying to accomplish here.

merittrompeter:
Thank you so much for the helpful comments @howarddierking  and @nicklathe!

I’ve tinkered with a jQuery solution and I seem to be running into the same two issues I did last week:

I can't seem to get any events delegated to the dynamically loaded swagger content to trigger. 

Once I get the event to trigger (by putting the listener on the container the swagger is loaded into instead of a swagger-generated elements) I need to delay replacing the text until it’s loaded.

So here’s something that works, but I wonder if you’ve any ideas that better address these two issues?

$('.col-md-9').on('click', function(){
  setTimeout(function(){
    $("pre").text(function () {
        return $(this).text().replace("developer.concur.com/api-explorer-proxy", "www.concursolutions.com");
  })
}, 500);
});

merittrompeter:
Thanks @nicklathe for his solution. We're now using .replace on the incoming url as well so we no longer have to scan the DOM.

swagger-ui/swagger-ui.js
nicklathe: @howarddierking Atom purposely auto strips trailing white space. This is by design, as their argument is that code shouldn't have trailing white space in the first place. So when a file is opened in Atom, it's going to go through and remove all the trailing white spaces. Makes for a messy PR at times, but in the end I'd imagine it's a good thing?

PR #448, opened 2016-03-17
Update launch-external-url.markdown

dstaab:


henryyamamoto-concur:
I think we should try to avoid ALL CAPS in our documentation and use **boldface** instead. To do this, add "**" at the beginning and end of where you want the boldface text.

Also, I think you were trying to add bullets. To make each sentence a bullet, add a "*" space, followed by the sentence.

Right now, the modified copy is showing up is a code block.

dstaab:
Ok, Changes committed.

PR #450, opened 2016-03-17
enable blog feed

merittrompeter:
Enabling RSS for subscribers.

Note: Once this is in I'll be looking into adding subscribe with buttons for popular readers.


nicklathe:
:shipit:

PR #456, opened 2016-03-18
local storage to preserve token

merittrompeter:
Part of the feedback we received on Swagger 2.0 was frustration at have to enter your API key repeatedly. We could use local storage to preserve it for the user.


nicklathe:
This is exactly what I was thinking.

nicklathe:
Awesome! :shipit:

PR #460, opened 2016-03-21
update 'reference'

merittrompeter:
Update 'reference' on blogpost to fix view.

https://developer.concur.com/blog/2016/03/18/quick-expense-api-payment-type.html


henryyamamoto-concur:
Thanks!

PR #466, opened 2016-03-23
update country value

merittrompeter:
Realized this form wasn't behaving properly after we changed the country data values to sync with marketing data - changed it to match.


henryyamamoto-concur:
Let's merge this sucker through to live if possible.

nicklathe:
:shipit:

PR #467, opened 2016-03-23
add Follow On Feedly button

merittrompeter:
Enable 'Follow On Feedly' button; add popover on hover to for user info.


nicklathe:
:shipit:

PR #469, opened 2016-03-24
use variable to replace Swagger URL

merittrompeter:
Switching the replacement URL in Swagger UI with variable determined by Jekyll site.data.branch.branchname.

### DIFF
--- a/_includes/swagger.html
+++ b/_includes/swagger.html
@@ -6,12 +6,17 @@
 <!-- <link href='/swagger-ui/css/ecebook.css' rel='stylesheet' type='text/css'/> -->
 
 <!-- <script src='/swagger-ui/lib/jquery-1.8.0.min.js' type='text/javascript'></script> -->
+
+  {% if site.data.brach.branchname ==  'preview' %}
+    {% assign URL = 'preview.developer.concur.com/api-explorer-proxy' %}
+  {% else %}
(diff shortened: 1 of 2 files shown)


_includes/swagger.html
```diff
@@ -6,6 +6,12 @@
 <!-- <link href='/swagger-ui/css/ecebook.css' rel='stylesheet' type='text/css'/> -->
 
 <!-- <script src='/swagger-ui/lib/jquery-1.8.0.min.js' type='text/javascript'></script> -->
+<script>
+  var branch = "{{site.data.branch.branchname}}";
+  var replacementURL;
+  (branch === 'preview') ? replacementURL = 'preview.developer.concur.com/api-explorer-proxy' : null;
+  (branch === 'livesite') ? replacementURL = 'developer.concur.com/api-explorer-proxy' : null;
+</script>
```
howarddierking: A couple of questions/thoughts here:

1. Did you consider doing the conditional code using liquid rather than in the javascript code?
1. Do we have a plan for what we do in the event that branch name is neither preview nor livesite? Setting `replacementurl` to null would, I imagine, have unexpected results downstream.

merittrompeter: Great thoughts, and thanks for the clarification!

nicklathe:
:shipit:

PR #471, opened 2016-03-28
enable tracking across forum

merittrompeter:
Setting up out GA to be able to track visitors across our forum.

There will also be script on the forum site:
https://forum.developer.concur.com/admin/customize/css_html/7/body-tag


nicklathe:
Looks good. :shipit:

PR #472, opened 2016-03-28
Update index.markdown

dstaab:


dstaab:
need to have an "s" in the http://concursolutions.com ref at the bottom of the page

PR #475, opened 2016-03-29
Update AttendeeTypes.markdown

jimandersen:
Henry,

Let me know if I need to include the Schema on this and other extract pulls. I'm missing data from my profile like a specific id besides my jim@concurconnect2.com log in. So I don't have the profile data I need at this point to successfully run the extracts. 

One other question, given the choice to run a json extract and an XML extract, which do you want. I am running json extracts but can easily replace them with XML.


henryyamamoto-concur:

PR #476, opened 2016-03-29
Update Allocations.markdown

jimandersen:
This should be good to go unless you want to add any proscriptions about using the ID from the GET Expense Allocations.


henryyamamoto-concur:

PR #477, opened 2016-03-29
Update Attendees.markdown

jimandersen:
I left place holders for successful XLM strings for the following:

Post Attendees: I need to have data I can use to configure the schema correctly. Or I get errors like Invalid attendee type ID 

Delete Attendees:
I got a server error when I used both the attendee Id and when I used my Admin credentials


henryyamamoto-concur:

PR #478, opened 2016-03-30
Update ConnectionRequests.markdown

jimandersen:
Placeholders inserted for XML Examples of successful responses for the following: 

	• GET Common connection requests
		○ Need a valid supplier ID
	• PUT common connections - create a connection on behalf of the specified user
		○ Need valid Supplier ID
	• DELETE common connection
		○ Need a valid specified resource to complete the process
	• GET Common connection requests by ID
		○ Need a resource with the specified ID
	• PUT Common connection requests by ID
		○ The schema is incorrectly configured because the data at the root level is invalid. Need a valid object ID

### DIFF
--- a/api-explorer/v3-0/ConnectionRequests.markdown
+++ b/api-explorer/v3-0/ConnectionRequests.markdown
@@ -3,7 +3,100 @@ title: Connection Requests
 layout: reference
 reference-type: swagger
 ---
+##Connection Requests
 
+###GET Common Connection Requests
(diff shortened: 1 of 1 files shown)


henryyamamoto-concur:
@jimandersen Are you cutting and replacing all the text? The only reason I ask is that I'm not seeing the diffs from what was there originally.

jimandersen:
Then I must be doing something wrong in GitHub because I'm building the pages using .md directly.

henryyamamoto-concur:

henryyamamoto-concur:

PR #480, opened 2016-03-30
Update DigitalTaxInvoices.markdown

jimandersen:
Digital Tax invoices
	• GET Digital Tax invoices by ID
		○ Need a resource with the specified digital tax ID
	• PUT Digital Tax Invoices by ID
		○ Need a list of tax invoice statuses to allow XML string to execute


henryyamamoto-concur:

PR #481, opened 2016-03-30
Update Entries.markdown

jimandersen:
Entries
	• PUT Expense entries
		○ Data at the root level is invalid - need expense entry data to use
	• DELETE Expense entries
		○ Need resource with the specified ID
	• GET Expense Entries by ID
		○ Need a resource with a specified resource
	• PUT Expense Entries by ID
	• Requires valid data at the root level: expense entry ID and data to configure the Model Schema

### DIFF
--- a/api-explorer/v3-0/Entries.markdown
+++ b/api-explorer/v3-0/Entries.markdown
@@ -4,6 +4,133 @@ layout: reference
 reference-type: swagger
 ---
 
+###Entries
 
+####GET Expense Entries
+This endpoint allows users to GET all expense entries owned by the user.
+
+
+####Parameters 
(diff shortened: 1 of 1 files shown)


jimandersen:
That is from the original file structure. I assume it supports the "try it out" function on each of the pages.  I left it there because I thought you would want to include a swagger test opportunity. I can scrape it out of the rest as I go along if you want.

henryyamamoto-concur:
It's not rendering correctly in preview.

jimandersen:
I haven't touched it other than to allow it to push down when I added more content. Are you saying the pull request is not rendering correctly or the swagger/api-explorer isn't?

henryyamamoto-concur:

PR #485, opened 2016-03-31
Update attendee-type-resource.markdown

jimandersen:
Post attendee types (Create a new attendee)
Need valid Attendee FormID. Using one from the ALL search does not work. Did not insert invalid string or Placeholder note.

Delete attendee type
Need a valid attendee ID to delete. Did not insert invalid string or Placeholder note. Do you want me to use carets instead of the attendee ID I pulled from the All search? 

Example: https://www.concursolutions.com/api/v3.0/expense/attendeetypes/gWjeLsVUzCo9GKsE78ZtoAME9 
to 
https://www.concursolutions.com/api/v3.0/expense/attendeetypes/<attendeeID>

PUT Attendee type by ID
Need valid Attendee FormID. Using one from the ALL search does not work. Did not insert invalid string or Placeholder note.


henryyamamoto-concur:
@dstaab We need to provide Jim with additional help.

Issue #491, opened 2016-04-04
Fix case for "ERP Integration"


henryyamamoto-concur:
Fix "Erp-integration" in left-hand nav of API Guides section to "ERP Integration"

PR #493, opened 2016-04-04
edit case in liquid

merittrompeter:
Issue #491.


nicklathe:
This looks good. The one-off thing for ERP feels wrong, but I don't see a way around it at this point. :shipit:!

PR #539, opened 2016-04-29
Update sandbox-configuration-assistance.html

elhansen748:
I don't know where this goes when a user fills it out?  I think this link should be removed from the dev portal, and partner developers should be instructed to Post to the forums if they need this type of assistance.

### DIFF
--- a/tools-support/sandbox-configuration-assistance.html
+++ b/tools-support/sandbox-configuration-assistance.html
@@ -1,3 +1,6 @@
+
+I suggest we remove this page and direct partner developers to use the forums instead
+
 ---
 layout: page
 title: Sandbox Configuration Assistance


henryyamamoto-concur:
This information goes to a Marketo database instead of the old Eloqua database. Did anyone ever go into Eloqua for this info before? If not, then we'll just provide a link to the forums. I'll also create a new forum category for "Sandbox Setup Assistance"

elhansen748:
I think we use to receive an email from Eloqua when this form was filled out.  We never went into Eloqua at all that I know of.   I think a link to the forum is the best way for them to get help.

Erik

henryyamamoto-concur:
Fixed this in preview. Will push out to livesite soon

PR #549, opened 2016-05-11
userId is not supported by this API

visadin:
Based on our investigation of this API, there is no code to support userId as a query string parameter to filter the results. Therefore, removing this from the swagger definitions.


henryyamamoto-concur:
@merittrompeter Can you take a quick look at this change and merge if you don't see any issues? Thanks.

PR #568, opened 2016-06-07
Update 1create-a-user.markdown

dstaab:


dstaab:
cleaned up some formatting with the bulleted points for consistency

henryyamamoto-concur:
Is that what you really want it to look like? The "a" level is running right after the parent with no indent or break.

dstaab:
The latest pull request was submitted to address all of that !  Does it not look like that?  I thought you said using spaces is how to manage an outline?

henryyamamoto-concur:
If you look at the example I provided in the other pull request, you'll see that the spaces are in front of the bullet or number.

In markdown sub bullets look like this:

## Bullets
* First level
 * Second level
   * Third level
     * Fourth level

## Numbers
1. Level one
  2. Level two
     3. Level three
        4. Level four

You don't need to put the actual numbers/letters in. They could all be "1"s.

Here's what your latest pull request is rending as:
screen shot 2016-06-07 at 10 49 47 pm

PR #596, opened 2016-07-05
Added an example section

cmartinezledo:
The reason of this commit, is to avoid forum questions about this endpoint. There are like 2 or 3 posts of the same issue. And there is no place in docs that says the xmlns of UserBatch. 
Thought that after my post someone would fix this issue.
Proof: https://forum.developer.concur.com/t/how-do-you-reset-an-user-password/332/10


cmartinezledo:
@henryyamamoto-concur is it good for you?

PR #608, opened 2016-07-19
Updating Itinerary codes

Trudeaucj:


henryyamamoto-concur:
Note that codes may vary from vendor to vendor

PR #640, opened 2016-07-26
Update payment-batches.markdown

dstaab:
Henry - I will cc you on an email about this change coming from Zach in Support.  He was requeseted a new flow order of sections within this page:

Retrieve a List of batches should come first

Close Batch should come 2nd

Retrieve actual file should be at the bottom.

I cut and pasted the sections as needed but will the linked table of contents get screwed up or will it auto-adjust?


henryyamamoto-concur:
Does the version section only apply to retrieve file? This section got moved as well so double checking before merging.

Version

1.1 for GET methods
1.2 for POST method

PR #641, opened 2016-07-26
Update 3how-to-identify-group-codes-for-standard-edition-clients.mark…

dstaab:
…down


bhague1281:
@dstaab Can you resolve the conflicts in this file, or make a new PR with your fixes?

PR #644, opened 2016-08-01
Version 1.0 is deprecated and updating request for version 1.2

Rilussion:


henryyamamoto-concur:
@Rilussion 
Are you actually deprecating v1.0 and directing users to v1.2? I'm seeing what's posted on the livesite is already v1.2 so are you just fixing an issue in the documentation?

If you are actually deprecating, you must notify clients and partners with 90 days notice prior to deprecation. Please see the [deprecation policy](https://developer.concur.com/tools-support/reference/deprecation-policy.html)

Rilussion:
I removed 'deprecated' verbiage. Also updated api with more examples.
Please review.

PR #647, opened 2016-08-04
Cleaning up general Callout info

jmtrachy:
Made three changes:

1) Added the two additional supported calls to the list of calls
2) Removed the section about Concur giving out a test Tomcat application
3) Added the test connection endpoint


henryyamamoto-concur:
Excellent. Thanks for updating this.

PR #657, opened 2016-08-23
Added vendor group and vendor bank to API spec

jtbigwoo:
Vendor group and vendor bank were added to the vendor API in the June release, but never added to the documentation.  Contact info: Tom Bigwood at concur


henryyamamoto-concur:
Thanks for the update. @jtbigwoo Can you also update the JSON that powers the API Explorer (swagger)?  https://github.com/concur/developer.concur.com/blob/preview/api-explorer/v3-0/Vendors.swagger2.json

PR #674, opened 2016-09-16
Adding separate vendor bank and vendor group APIs

jtbigwoo:
Also fixed a couple typos.


henryyamamoto-concur:
Merged into preview. Also, when you have a moment, can you update the explorer as well?

https://github.com/concur/developer.concur.com/wiki/Update-Swagger

PR #677, opened 2016-09-21
add change log

merittrompeter:
Enabling a change log section for notifications regarding API changes.

To put  post into change log vs putting something in blog add these two lines to the frontmatter:

categories:
  - changelog

Then drop in the _posts/changelog folder.


nicklathe:
:shipit:

PR #680, opened 2016-09-22
change auth header scheme based on token format

howarddierking:
looks at the token to determine the scheme for the authorization header
* if the token starts with a number and underscore (e.g. '1_...'), set scheme to `OAuth`
* otherwise, set scheme to `Bearer`

### DIFF
--- a/_includes/swagger.html
+++ b/_includes/swagger.html
@@ -30,13 +30,15 @@
 
 <script type="text/javascript">
   var swaggerUI = undefined;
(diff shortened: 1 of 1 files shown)


henryyamamoto-concur:
@nicklathe and @merittrompeter Take a look?

nicklathe:
Looks good to me. :shipit:. .test is so cool. Also Henry, you've got a new dev on dev center. I think Howard was talking to @bhague1281 about taking a look at this.

PR #689, opened 2016-09-28
replaced proxy url with connect url

bhague1281:
Removed code that relied on proxy URLs, replaced hosts with "us.api.concursolutions.com" and base paths with "/connect/"


howarddierking:
:shipit:

Issue #659, opened 2016-08-24
Update content on Why Concur page


henryyamamoto-concur:
From: @cskelley 

On https://developer.concur.com/why-concur.html, combine the “Developers” and “Partners” section into a new section titled “Independent Software Developers and App Center Partners”. Let me know if that title is too long. You can use whichever graphic/photo you prefer. 

**Use the following copy:**
 

Tap the $1.3 trillion business travel market* and deliver solutions to integrate easily with our APIs for Travel, Expense and Invoice. Promote your services through the Concur App Center, featuring apps such as Uber, Starbucks, Airbnb and Starwood.
 
Hear what our customers and partners are saying about the value of Concur integrations:
 
{insert current “Watch Videos” link and graphic here}
 
We’re looking for qualified ISVs and third party developers with compelling integration ideas for Concur’s 40 million business travelers and 33,000 enterprises worldwide. If your business is well established and you’re committed to develop, market and support your integration, then we’d like to talk with you about our Platform Partner Program (fees apply). Click the link below to send an email to our Platform & Partner Solutions business development team.
 
{insert current “Contact us” link and graphic here}
 
*GBTA Foundation

cskelley:
Hey @henryyamamoto-concur: Any update on when this might be completed?

henryyamamoto-concur:
@cskelley Please review: https://preview.developer.concur.com/why-concur.html

cskelley:
@henryyamamoto-concur It looks great!

PR #700, opened 2016-10-04
Consolidated/updated content on Why Concur page

bhague1281:
#659


henryyamamoto-concur:
Merging so I can send to Chris Kelly for review and approval.

Issue #529, opened 2016-04-25
Add print function to Terms of Use


cskelley:
https://developer.concur.com/Terms-of-Use.html

When I'm negotiation with prospective new Platform partners, they often want a copy of the API License and Services Agreement (our agreements refer to this URL). It would be nice if there were a print function on this page so the partner could print out the API License and Services Agreement without all the other text that appears on the page.

henryyamamoto-concur:
Add a print button that only prints the terms content and not the navigation and other extraneous elements.

PR #719, opened 2016-10-12
print button, hide navigation when printing pages

bhague1281:
#529 Print button added to the top of the "Terms of Use" page only. Added a Bootstrap class to hide navigation components when printing any page. @howarddierking


nicklathe:
I didn't run it locally, but code looks cool. It gets a :shipit: from me.

PR #748, opened 2016-10-28
removed sandbox assistance form from left nav

bhague1281:
...by moving the page out of the tools and support directory


bhague1281:
#748

PR #769, opened 2016-11-10
added company level auth docs

lokeuei:
- added docs
- added diagram


henryyamamoto-concur:
This is live now, so please update here as well: https://github.com/concur/developer.concur.com/tree/preview/api-reference/authentication

henryyamamoto-concur:
Or just update the live and leave preview alone.

PR #822, opened 2016-12-13
expense entry post doc

bhague1281:
Links need to be changed. @henryyamamoto-concur


henryyamamoto-concur:
My bad. This should be in v1.1 folder. I was looking at the URL and it says 1.1

PR #955, opened 2017-03-11
Update register.html

dstaab:


manage-apps/register.html
```diff
@@ -15,9 +15,11 @@
    <p>    
     If you are unsure of your technical contact, you can reach out on the Concur Developer <a href="https://forum.developer.concur.com/">
     Forums</a> 
-    and someone will be able to help identify the right resource for you.  Please indicate in your forum Post whether you are
-    looking to create an application for all Concur clients to potentially use, are developing for a specific client, or if you are a 
-    TMC/reseller.
+    and someone will be able to help identify the right resource for you.  Indicate in your forum Post one of these 3 goals:
+
+     1. develop an application for the Concur App Center (requires a Partner Agreement)
```
bhague1281: This file is a `.html` file, so you'll need to use HTML tags to format these as list items. Like so:

```html
<ul>
  <li>1. develop an application...</li>
  <li>2. develop an application for a specific...</li>
</ul>
```

PR #971, opened 2017-03-28
Update 01-profile-resource.markdown

mmessa:


bhague1281:
Can we get some context for this PR?

mmessa:
Adding Sabre Profile Id to travel profile API2.0, for both update and get.

bhague1281:
Thanks! Made a minor change to prevent a break in format :shipit:

PR #986, opened 2017-04-04
Add description to localized data swagger

elhansen748:


api-explorer/v3-0/LocalizedData.swagger2.json
```diff
@@ -15,7 +15,7 @@
   ],
   "info": {
     "title": "Localized Data",
-    "description": "",
+    "description": "Invoice includes various status codes, and error messages. This API provides a method to list the localized strings of various codes.',
```
bhague1281: Double quote needed instead of a single quote at the end of this line.

PR #1006, opened 2017-04-06
adds detail to receipts date-time format descriptions

cfreiner:
This adds a more detailed description to the date-time format we use in receipts schemas. The schema docs script seems to have added some other descriptions that weren't previously in the docs as well.

Question: our schem doc generation script seems to spit out different numbers of `#` symbols compared to what actually appears in the docs. @pranali-concur Were these edited by hand? I thought that the schema docs were just copy-pasted from the script output.

### DIFF
--- a/api-reference/receipts/supported-receipt-types.markdown
+++ b/api-reference/receipts/supported-receipt-types.markdown
@@ -20,7 +20,7 @@ See the schema documentation below for the specifications of each type, plus the
 
 ### Schemas
 
-#### Address-Original
+## Address-Original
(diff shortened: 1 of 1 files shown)


bhague1281:
It looks like @henryyamamoto-concur was the one who added the initial schema docs. If possible, we should adhere to using 4 `#` symbols instead of 2, in order to denote that each schema is a subtopic of the **Schemas** topic. This is done for SEO + accessibility.

cfreiner:
Got it, `#`'s adjusted to 4 for the main schemas and 5 for their children.

PR #1009, opened 2017-04-07
Update expense-report-get.markdown

dstaab:
This API is not officially deprecated and it cannot be until v4 is released.  note: the v3 Entries API should not be used due to a major issue with the response related to reports being created by a Delegate - the report content does not come back in the response.  So, when we refer devs to this page, they instantly ask about the deprecation reference.  Can this be moved to a page where it doesn't state "deprecated" ?


dstaab:
This pull request is larger than the simple period I added per the above comment.  I couldn't find any other way to convey the need for this change since the action buttons didn't get enabled untnil an actual change was made.

bhague1281:
See #1010

PR #1014, opened 2017-04-13
Update revoke token doc

elhansen748:
The revoke token endpoint changed fromhttps://api.concursolutions.com/appmgmt/v0/connections to https://us-api.concursolutions.com/appmgmt/v0/connections but the documentation wasn't updated to reflect this.


bhague1281:
Just to clarify, this shouldn't be `us.api.concursolutions.com`?

elhansen748:
This should be https://us-api.concursolutions.com/appmgmt/v0/connections now, not https://api.concursolutions.com/appmgmt/v0/connections

PR #1031, opened 2017-05-02
Adding table to make deprecation policy clearer

elhansen748:
Loke requested we make this change, as developers were confused about the deprecation policy.


bhague1281:
LGTM :shipit:

PR #1122, opened 2017-07-26
Fixing broken links left over from migration to new DevCenter

bhague1281:


elhansen748:
Hey Brian, wow that is a lot.  Changes look good to me, and should clean things up nicely.

PR #1126, opened 2017-07-29
Clarifying the user of Geolocations in advance of the China DC roll out

brittanywest:
All API users should leverage the geolocation to ensure that they are utilizing the appropriate endpoints for their apps. Adjusted the docs for base URI to make that clear.

### DIFF
--- a/api-reference/authentication/apidoc.markdown
+++ b/api-reference/authentication/apidoc.markdown
@@ -15,7 +15,7 @@ If you are an existing partner with an existing app, please read both the [Migra
   * [Refreshing a token](#refresh_token)
   * [Revoking a token](#revoke_token)
   * [Token Management](#manage_token)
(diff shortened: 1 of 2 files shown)


api-reference/authentication/apidoc.markdown
```diff
@@ -201,7 +201,10 @@ e013335d-b4ce-4c43-a7e4-b67abc1adcb0
 
 It is highly recommended that you store Refresh Tokens together with your user's authorization metadata in your application every time you obtain a new `refreshToken` as they might change depending on different scenarios.
 
-## <a name="base_uri"></a>Base URIs
+## <a name="
+_uri"></a>Base URIs
```
bhague1281: Extra newline character here, which breaks the link. Otherwise 👍

brittanywest:
Thx updated.

PR #1137, opened 2017-08-10
Update apidoc.markdown

sheylat:


elhansen748:
Spoke with Sheyla.  i will work with Loke to get the text she has requested added.

PR #1162, opened 2017-09-15
Replace enablement guide

brittanywest:
Replacing the legacy enablement guide with a new guide and updating references.


brittanywest:
@dstaab For review

PR #1169, opened 2017-09-22
Adding docs for multiple geolocations

brittanywest:
This updates the Base URI and "obtaining a token" to support US & EU and clarify geolocation usage.


brittanywest:
@dstaab for review

PR #1190, opened 2017-10-11
Updating the App Center client FAQ

brittanywest:


brittanywest:
Oops - should have been pdf. Updated.

PR #1192, opened 2017-10-15
Update apidoc.markdown

dstaab:
can you also please bold the following text and use a different color such as blue?
"When obtaining the token, the token's geolocation will be included in the response. The token's geolocation should be stored along with the token. The Developer's app will then be able to make subsequent calls using the correct end points per GEO location."


bhague1281:
I put the text within a blockquote and bolded it. Let me know if that will suffice. I'd prefer to avoid using different colors if possible.

dstaab:
that works!  When is this scheduled to go to production?  No hurry, just anxious to get some of this old auth and old sandbox stuff updated.


Doug Staab   |   Director, App Center Partner Enablement    |   425.590.5929

________________________________

bhague1281:
If everything on preview looks ok, I can push today.

PR #1211, opened 2017-10-20
Update old & new auth comparison diagram 

zenan-angus-wang:
I changed the client name in the graph to be generic.


api-reference/authentication/oauth2-migration-best-practices.markdown
```diff
@@ -73,6 +73,6 @@ layout: reference
     - for a full list, review this doc: https://developer.concur.com/api-reference/authentication/apidoc.html#response_codes
 
 ### 4. Old auth v.s. new auth diagram
-![old v.s. new](https://raw.githubusercontent.com/concur/developer.concur.com/preview/api-reference/authentication/newAndOldAuthComparison.png)
+![old v.s. new](https://raw.githubusercontent.com/zenan-angus-wang/developer.concur.com/preview/api-reference/authentication/oldNewAuthComparion.png)
```
bhague1281: I should've caught this in the last PR, but you can reference the image using a relative path, which is preferred.

```
![old vs. new](./oldNewAuthComparion.png)
```

zenan-angus-wang:
changed to relative path

PR #1253, opened 2017-11-20
Uploading Images for new Receipts Recipe

sheylat:
This Pull Request is for uploading images that will be used within the new Receipts Recipe page under API Guides.  (Update e-receipts.markdown #1255)

appflow_listing

webflow_auth

signin

otl email2

otpflow

otp emaiil

error2


bhague1281:
Images should reside in the repository so they can be kept track of and be hosted from `developer.concur.com`. I'll add these images to the repository in a separate PR.

Issue #1231, opened 2017-11-02
App creation tool needs to support "legacyId"


elhansen748:
@elhansen748 commented on [Mon Apr 10 2017](https://github.com/concur/developer.concur.com-forms/issues/121)

In order to call some old connect APIs, when an an app is created a "legacyID" needs to be set.  the legacyID is the Outtask Company ID of the site that is needed for the old API calls to function.  A legacyID should also be set if a partner has an existing old-auth app and is creating a new auth app, so that we know the same partner owns both.

- [ ] A legacyID field should be added to the app creation form.  
- [ ] The legacyID field should not be required.
- [ ] No validation of the id should be done.  Assumption is the user will know and enter the correct value.  If they enter the wrong value they can edit the app and fix it.
- [ ] When a user logs in to view all their apps, if a legacyID is returned it should be displayed to the user in the UI.
- [ ] The user should be able to edit the legacyID of any app they own.

Sample of legacy id:

"com:concur:internal:product:Identifiers:1.0": {
    "legacyId": "12076"

princemoseslive:
Fix changes pushed into the repo

PR #1267, opened 2017-12-06
Issue id 1231

princemoseslive:
PR Against this issue url:
https://github.com/concur/developer.concur.com/issues/1231 

Requirements:

 A legacyID field should be added to the app creation form.
 The legacyID field should not be required.
 No validation of the id should be done. Assumption is the user will know and enter the correct value. If they enter the wrong value they can edit the app and fix it.
 When a user logs in to view all their apps, if a legacyID is returned it should be displayed to the user in the UI.
 The user should be able to edit the legacyID of any app they own.

Issue Status : Resolved

### DIFF
--- a/app-management/components/AppDetailsPage/AppEditForm.jsx
+++ b/app-management/components/AppDetailsPage/AppEditForm.jsx
@@ -32,11 +32,13 @@ const constraints = {
   appType: {
     presence: { message: 'is required' },
   },
+  LegacyId: {
+  },
   redirectUris: {
     validateUrlArray: {
       allowLocal: true,
     },
-    firstElementRequired: false,
+    firstElementRequired: true,
   },
   allowedGrants: {
     presence: { message: '- at least one is required.' },
@@ -79,6 +81,15 @@ const EditAppForm = ({ handleSubmit }) => (
             label="App Type"
             options={appTypes}
             simpleValue
+          />
+        </section>
+        <section className="col-md-8">
+          <Field
+            component={InputField}
+            type="text"
+            name="legacyId"
+            label="Legacy Id &#42;"
+            placeholder="Legacy Id"
           />
         </section>
         <section className="col-md-8">
--- a/app-management/components/NewAppPage/NewAppForm.jsx
+++ b/app-management/components/NewAppPage/NewAppForm.jsx
@@ -34,11 +34,13 @@ const constraints = {
   appType: {
     presence: { message: 'is required' },
   },
+  LegacyId: {
+  },
   redirectUris: {
     validateUrlArray: {
       allowLocal: true,
     },
-    firstElementRequired: false,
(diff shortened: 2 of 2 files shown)


app-management/components/AppDetailsPage/AppEditForm.jsx
```diff
@@ -32,11 +32,13 @@ const constraints = {
   appType: {
     presence: { message: 'is required' },
   },
+  LegacyId: {
```
bhague1281: We can leave this out, since no constraints are needed for this field.

app-management/components/NewAppPage/NewAppForm.jsx
```diff
@@ -34,11 +34,13 @@ const constraints = {
   appType: {
     presence: { message: 'is required' },
   },
+  LegacyId: {
```
bhague1281: See above

app-management/components/NewAppPage/NewAppForm.jsx
```diff
@@ -34,11 +34,13 @@ const constraints = {
...
+  },
   redirectUris: {
     validateUrlArray: {
       allowLocal: true,
     },
-    firstElementRequired: false,
+    firstElementRequired: true,
```
bhague1281: See above

app-management/components/AppDetailsPage/AppEditForm.jsx
```diff
@@ -32,11 +32,13 @@ const constraints = {
...
+  },
   redirectUris: {
     validateUrlArray: {
       allowLocal: true,
     },
-    firstElementRequired: false,
+    firstElementRequired: true,
```
bhague1281: This should be false. We're not requiring redirectUris anymore.

app-management/components/AppDetailsPage/AppEditForm.jsx
```diff
@@ -81,6 +83,15 @@ const EditAppForm = ({ handleSubmit }) => (
             simpleValue
           />
         </section>
+        <section className="col-md-8">
+          <Field
+            component={InputField}
+            type="text"
+            name="legacyId"
+            label="Legacy Id &#42;"
```
bhague1281: Can we change this label to just "Legacy Id"? The `&#42;` is only needed to denote that the field is required.

app-management/components/NewAppPage/NewAppForm.jsx
```diff
@@ -90,6 +92,15 @@ const NewAppForm = ({ handleSubmit }) => (
             options={appTypes}
           />
         </section>
+        <section className="col-md-8">
+          <Field
+            component={InputField}
+            type="text"
+            name="legacyId"
+            label="Legacy Id &#42;"
```
bhague1281: See above

PR #1350, opened 2018-03-09
Adding v4 quick expense service to api-explorer

patilanand:
Adding v4 quick expense service swagger documentation (api-explorer)to developer portal.

We would like to verify our changes first in preview.developer.concur.com before we push to production.


patilanand:
I have verified that the v4 quick expense API works on preview. Please merge this to livesite.
pasted image at 2018_03_23 10_47 am

PR #1422, opened 2018-05-25
Update index.markdown

dstaab:
## Pull Request Checklist

- [ ] Proofread documentation changes for spelling and grammatical errors
- [ ] Used Markdown code blocks for all applicable examples using code
- [ ] Used relative links when linking to other documentation on Developer Center
- [ ] If creating new API reference pages, followed the guidelines stated [in the wiki](https://github.com/concur/developer.concur.com/wiki/Creating-API-Reference-pages)

### DIFF
--- a/src/api-reference/index.markdown
+++ b/src/api-reference/index.markdown
@@ -8,34 +8,9 @@ redirect_from:
 
 # Quick Start Guide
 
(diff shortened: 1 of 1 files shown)


dstaab:
removing the sandbox steps since they cause confusion for web service customers because the web services team does not use sandboxes.  The Certification Team (my team) is already in contact with the partners at this point, so we have already provisioned sandboxes for those partners.  The removed text, therefore, no longer applies to either one of our business models (Web Services integrations or App Center Partner integrations)

PR #1429, opened 2018-06-01
TVR-8225-Developer Center - Request API v4 documentation - Documentat…

christellethiebaut:
…ion correction

Change the preview to fix some errors

## Pull Request Checklist

- [x] Proofread documentation changes for spelling and grammatical errors
- [x] Used Markdown code blocks for all applicable examples using code
- [x] Used relative links when linking to other documentation on Developer Center
- [ ] If creating new API reference pages, followed the guidelines stated [in the wiki](https://github.com/concur/developer.concur.com/wiki/Creating-API-Reference-pages)

### DIFF
--- a/src/api-reference/request/v4.request.markdown
+++ b/src/api-reference/request/v4.request.markdown
@@ -3,53 +3,48 @@ title: Request endpoints
 layout: reference
 ---
 
-# Request endpoints
+# <a name="request-endpoints"> Request Endpoints
 
 ** Pre-release Version 4.0**
 
 If a new or existing project please use version [3.0](./v3.request.html) of this API.
 
-* [Resources](#resource)
-  * [Request](#request-resource)
-    * [Create a new Request](#request-resource-create)
-    * [Get the content of an existing request](#request-resource-get)
-    * [Update the content of an existing Request](#request-resource-update)
-    * [Delete an existing Request](#request-resource-delete)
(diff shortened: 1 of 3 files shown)


src/api-reference/request/v4.request.markdown
```diff
@@ -88,6 +84,8 @@ Name|Type|Format|Description
 ---|---|---|---
 -|object|[Request](#request)|The Request having {requestUuid} as unique identifer
 
+[Back to top](#request-endpoints)
```
charlieconcur: Let's get rid of these 'Back to top' links -- they unnecessarily complicate the documentation and with anchor points the back button in the browser serves this purpose and these tend to interfere with that muscle memory as well.

christellethiebaut: Ok

cbobrie: Hi Charlie,
I got this idea from the Receipt Endpoing available in LiveSite from which we can take example from (https://developer.concur.com/api-reference/receipts/endpoints.html) and I feel that from an end-user experience using our documentation, this will avoid quantities of scrolling and make them access quicker to the required information.
Is this a blocking issue? Because I really would like to maintain those as from my perspective, I've been reviewing the documentation with and without those Back to top, and talking around me this is an added value to end-users. 
Thanks

PR #1442, opened 2018-06-12
Add docs for realtimeingest location API

SamarthaV-concur:
## Pull Request Checklist

- [x] Proofread documentation changes for spelling and grammatical errors
- [x] Used Markdown code blocks for all applicable examples using code
- [x] Used relative links when linking to other documentation on Developer Center
- [x] If creating new API reference pages, followed the guidelines stated [in the wiki](https://github.com/concur/developer.concur.com/wiki/Creating-API-Reference-pages)

### DIFF
--- a/src/api-explorer/v3-0/RealtimeingestLocation.markdown
+++ b/src/api-explorer/v3-0/RealtimeingestLocation.markdown
@@ -1,0 +1,7 @@
+---
+title: Realtime Ingest Location
+layout: reference
+reference-type: swagger
+---
+
+{% swagger /api-explorer/v3-0/RealtimeingestLocation.swagger2.json %}
--- a/src/api-explorer/v3-0/RealtimeingestLocation.swagger2.json
+++ b/src/api-explorer/v3-0/RealtimeingestLocation.swagger2.json
@@ -1,0 +1,197 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "description": "This API provides an endpoint to ingest real time user location information from Rideshare Services",
+    "version": "1.0",
+    "title": "Real Time Location Ingestion Service",
+    "contact": {
+      "email": "DL_AE4E8B60807542848C5FE406400E04C2@sap.com"
+    }
+  },
+  "basePath": "/realtimeingest",
+  "schemes": [
+    "https"
+  ],
+  "securityDefinitions": {
+    "Bearer": {
+      "description": "To access the API a valid JWT token must be passed in the Authorization header",
+      "type": "apiKey",
+      "name": "Authorization",
+      "in": "header"
+    }
+  },
+  "paths": {
+    "/location/:uuid": {
+      "post": {
+        "tags": [
+          "realtime-location"
+        ],
+        "summary": "Ingest real time location of users ",
+        "description": "",
+        "consumes": [
+          "application/json"
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "parameters": [
+          {
+            "in": "body",
+            "name": "body",
+            "description": "location object",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/PostObject"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "POST successful",
+            "schema": {
+              "$ref": "#/definitions/ApiResponse"
+            }
+          },
+          "401": {
+            "description": "Authentication Error",
+            "schema": {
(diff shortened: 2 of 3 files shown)


charlieconcur:
* Please base the documentation on this template: https://github.concur.com/developer/api/blob/master/documentation-template.md
* "This API supports POST request only." <-- Not necessary as this should be obvious from the documentation.
* Remove the **## Prerequisites** section entirely. It is referring to sandbox and is redundant (and yes, we should have other teams remove from their legacy documentation).
* Remove the **#### Retrieve a User Access Token** section entirely -- it is redundant as well to existing documentation in the Authorization section already present.
* The preceding underscore is messing up the formatting here: `_POST /location/:uuid`. Not sure whey this is immediately below the ## Schema section.
* Clean up all tables to match this formatting:

```
Property Name|Type|Format|Description
---|---|---|---
uuid|String|-|**Required** Contains the uuid of the user.
dropOffDateTime|String|DateTime|**Required** Date Time where the user was dropped off (in RFC3339 format)
dropOffLocation*|Object|JSON|**Required** Location where the user was dropped off
```

* Suggest you follow https://github.concur.com/developer/api/blob/master/guidelines.md#errors-when-4xx for error conditions.
* I'm not sure what the `:uuid` part of the URI is -- if it's a variable it should follow https://tools.ietf.org/html/rfc6570. I will assume it is.
* Separate the request from the payload and make it generic. Also, use the JSON keyword to discover bugs in samples (you are missing a comma at the end of `streetAddress`).

```
POST https://{baseURI}/location/{uuid}
Content-Type: application/json
Authorization: Bearer {access-token}
```

```json
{
  "uuid": "uuid",
  "dropOffDateTime": "2018-06-08T09:00:45-0600",
  "dropOffLocation": {
    "latitude": 47.610378,
    "longitude": -122.200676,
    "name": "Bellevue",
    "address": {
      "streetAddress": "601 108th Ave NE",
      "addressCountry": "US",
      "addressLocality": "Bellevue",
      "addressRegion": "WA",
      "postalCode": "98004"
    },
  }
}
```

* Response headers aren't in the form of JSON -- so that needs to be fixed.
* Watch for casing: `concur-correlationId` --> `concur-correlationid`.
* `concur-correlationid` is a header, not in the payload.
* Probably don't need `"message": "Event Received"` if `200 OK` is present unless you are planning to overload `POST` which is a different conversation.
* `uuid` is in the URI and in the payload -- it only needs to be in one place, I suspect.

SamarthaV-concur:
Thank you very much for the detailed review along with examples. I have tried my best to fix things to conform to the template and guidelines:-
* Please base the documentation on this template - Done
* "This API supports POST request only." <-- Not necessary as this should be obvious from the documentation. - Removed
* Remove the ## Prerequisites section entirely. It is referring to sandbox and is redundant (and yes, we should have other teams remove from their legacy documentation). - Removed
* Remove the #### Retrieve a User Access Token section entirely -- it is redundant as well to existing documentation in the Authorization section already present. - Removed
* The preceding underscore is messing up the formatting here: _POST /location/:uuid. Not sure whey this is immediately below the ## Schema section. - Removed
* Clean up all tables to match this formatting - Done
* Suggest you follow https://github.concur.com/developer/api/blob/master/guidelines.md#errors-when-4xx for error conditions. - Unfortunately, the API in it's current form does not conform to the error conditions guidelines. We currently return all the errors as a JSON Object {"fieldName": [Errors associated with the fieldName]}. I have a raised a Jira request to fix this with a version update on the API.
* I'm not sure what the :uuid part of the URI is -- if it's a variable it should follow https://tools.ietf.org/html/rfc6570. - Done
* Separate the request from the payload and make it generic - Done
* Response headers aren't in the form of JSON - Done
* `concur-correlationid` is a header, not in the payload - Our current API contract has this both in the response header and response payload. Will note this change for the API update
* Probably don't need `message`: `Event Received` if 200 OK - `message` field gets overloaded to indicate errors as well. Please let me know if there is a better way to do this.
* `uuid` is in the URI and in the payload - Will note this in the API update

charlieconcur:
@SamarthaV-concur Excellent job taking in the feedback. We still had some formatting challenges and I just found it easier to fix them for you -- see https://github.concur.com/developer/scratch/blob/master/realtime-ingest-location/temp.md for the source and then, if you agree, go ahead an integrate with your change.

SamarthaV-concur:
@charlieconcur - Thanks a lot. Appreciate your help. I have incorporated the formatting updates and pushed in a commit.

PR #1475, opened 2018-07-25
Add Pre-Release Text

haydenlam:
To the top of the page.


charlieconcur:
I'm not thrilled with the idea of 'Warning' so let's work together to refine the wording we see within https://github.concur.com/developer/api/blob/master/documentation-template.md (look at the example therein) and then bring it back to this pull request.

haydenlam:
Updated Prefix of the Pre-Release text to display `Preview` instead of `Warning`

PR #1478, opened 2018-07-27
trying to change the format of budget adjustments and see if that fixes the issue with the page not rendering

anuvenkataraman:
trying to change the format of budget adjustments and see if that fixes the issue with the page not rendering


charlieconcur:
Please make all of the tables use the following format and remove all of the extraneous spaces, including leading spaces in some cases:

```
Name|Type|Format|Description
---|---|---|---
`adjustments`|`Array[budgetAdjustment]`|`body`|**Required** A JSON representation of a array of Adjustments
```

PR #1482, opened 2018-07-30
Update endpoints.markdown

stepancannuscio:
Here are a few recommended changes which make it clearer how to post eReceipts/Image-Only Receipts

### DIFF
--- a/src/api-reference/receipts/endpoints.markdown
+++ b/src/api-reference/receipts/endpoints.markdown
@@ -325,6 +325,11 @@ Creating a receipt requires JSON data about the transaction and, optionally, an 
 
 Successful POST requests will receive a response of 201 Created. The `Location` header of the response contains a URL for your receipt. Once the receipt has been processed, it can be retrieved at this URL. The `Link` header of the response contains a processing-status URL for your receipt. More information can be found [here](#endpoint-get-receipt-status).
 
+Helpful Notes:
+- Include link as a header and make its value: “<http://schema.concursolutions.com/{receipt type}.schema.json>;rel=describedBy” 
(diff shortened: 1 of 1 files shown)


charlieconcur:
Please have a peer review done. I think @jolson88 or @kamsbry or their delegates can provide.

src/api-reference/receipts/endpoints.markdown
```diff
@@ -526,7 +531,9 @@ http https://us.api.concursolutions.com/receipts/v4/{RECEIPT ID}/image "Authoriz
 
 Successful POST requests will receive a response of 202 Accepted. The Location header of the response contains a URL for your receipt image. Once the receipt has been processed, it can be retrieved at this URL. The Link header of the response contains a processing-status URL for your receipt image.
 
-To post a receipt image, use multipart form data. The Content-Type:multipart/form-data header must be set. The image should be included under the key image.
+Helfpul Notes:
+- The header must include content-type with multipart/form-data as its value 
+- In the body, add "image" as a key and select "file" from the dropdown since you will be linking an image file. Then, choose your saved image file as the value.
```
jolson88: What dropdown are you referring to? As it's an API, talking about a dropdown feels confusing. Perhaps I'm just missing it out of context in this review :).

jolson88 (changes requested):
Just one minor knit-pick question to clarify. Otherwise, these changes look good to me :). Thanks for the changes :D.

src/api-reference/receipts/endpoints.markdown
stepancannuscio: I should have specified - I am talking about the context when running these calls through Postman. When doing so, there is a drop down in the body section where you can choose if the key is a text or file.

charlieconcur: We shouldn't be referencing any specific development tools in our documentation -- it should remain agnostic to developer tool choice.

PR #1485, opened 2018-08-01
Updating the AuthZ scopes page for accuracy.

charlieconcur:
This pull request makes the table markup simpler, clarifies some concepts and adds the APIs where found in the documentation.


cliffstrom:
Based on email thread, we don't want to put the API links on the same page, right?  Or do you think these changes are still good for developers to get a reference to the API?  Either way, we wouldn't be able to go live without more of the API links filled out.

charlieconcur:
I think you want scopes listed in the individual APIs and in a list form such as this one. I think these changes are good and should go live immediately even though we have some gaps to fill.

cliffstrom (changes requested):
Let's wait to make these updates until we have more data.

cliffstrom (approved):

PR #1487, opened 2018-08-03
Update the example of Password update

shinyay-jptc:
Not able to update the password by using current example as below.

But the request actually succeeds if we use "user" tab instead of "userprofile".

### DIFF
--- a/src/api-reference/user/index.markdown
+++ b/src/api-reference/user/index.markdown
@@ -211,9 +211,9 @@ Name | Type | Format | Description
 
 ```xml
 <UserBatch xmlns="http://www.concursolutions.com/api/user/2011/02">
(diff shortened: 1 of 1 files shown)


charlieconcur:
Please have a peer review done.

charlieconcur:
@bing-hu-concur Please take a look.

charlieconcur:
Ping @bing-hu-concur.

squarebing:
Bing Hu [4:26 PM]
have you verified the PR is correct?

Shinya Yoshimura [4:41 PM]
We Concur Japan team tested before submit PR.

squarebing:
Based on Shinya Yoshimura's confirmation, please approve

PR #1515, opened 2018-08-30
Move to Circle 2.0

meritmalling:
Thanks to @howarddierking for letting me reuse the code from his much cleaner yml file.

Please let me know anything that feels superfluous or clumsy -- or anything that could be improved.


.circleci/config.yml
```diff
@@ -0,0 +1,63 @@
...
+  build:
+    working_directory: ~/concur/developer.concur.com
+    docker:
+      - image: circleci/ruby:2.5.1
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - devcenter-deps-{{ .Branch }}-{{ checksum "Gemfile" }}
+          - devcenter-deps-preview-
```
howarddierking: what's this for?

.circleci/config.yml
```diff
@@ -0,0 +1,63 @@
...
+jobs:
+  build:
+    working_directory: ~/concur/developer.concur.com
+    docker:
+      - image: circleci/ruby:2.5.1
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - devcenter-deps-{{ .Branch }}-{{ checksum "Gemfile" }}
```
howarddierking: Oh, this is cool! I hadn't looked into using the cache yet, but I would imagine this will dramatically speed up build times.

howarddierking (approved):
A couple other questions.

* have we tested this, or will preview be where we test?
* Are the new AWS credentials setup in Circle?

merittrompeter:
Yes! I this was tested in the test-developer-concur-com bucket.

And the new credentials are indeed in the UI - also the old ones have been removed because on the first run they over-wrote the new ones.

.circleci/config.yml
merittrompeter: This will try to use a cache from your current branch first but default to a cache created from the preview branch if nothing has been cached on your current branch yet.

PR #1516, opened 2018-09-04
Updated URL

mahimarpi:


charlieconcur:
Merging because this is pre-release noting the new path would seem to be in conflict with this guideline: https://github.concur.com/developer/api/blob/master/guidelines.md#resource-naming.

PR #1520, opened 2018-09-05
Creating travel search event doc

brittnwestSAP:
Needs peer review by Valter and Barclay. TBD scope for this event and requires link to common event subscription documentation.


src/api-reference/common/event-pubsub/travel-search-event.markdown
```diff
@@ -0,0 +1,89 @@
+---
+title: Travel Search Event
+layout: reference
+---
+
+# Travel Search Event
```
charlieconcur: This is redundant to the Jekyll front matter above. Please remove.

PR #1521, opened 2018-09-05
TVR-8626

laurentlhermine:
New documentation from TVR-8626
Remove back to top links
Precision about the userId parameter
Get list of requests endpoint and example
New response codes


src/api-reference/request/v4.request.markdown
```diff
@@ -1,4 +1,4 @@
----
+# Ne---
```
charlieconcur: What is this...?

src/api-reference/request/v4.examples.markdown
```diff
@@ -148,20 +149,103 @@ Responde Body:
...
     }
   ]
 }
 ```
 [Back to top](#request-examples)
 
+#### <a name="request-resource-get-list"></a>Get the list of existing Requests
+#### GET /v4/requests
+```
+curl -X GET https://us.api.concursolutions.com/travelrequest/v4/requests?view=ALL&limit=10&start=0 --header 'Accept: application/json' --header 'Authorization: Bearer {token}'
```
charlieconcur: We need to remove references to specific tools. Follow the example given in the [Documentation Template](https://github.com/concur/developer.concur.com/blob/preview/src/document-template.markdown).

src/api-reference/request/v4.request.markdown
cbobrie: A bad copy paste from the markdown editor page => --- must be kept

src/api-reference/request/v4.examples.markdown
laurentlhermine: Is this really required ? curl is a pretty standard cross platform tool. I think it really helps developers to figure out what a valid call is.

src/api-reference/request/v4.request.markdown
laurentlhermine: corrected

src/api-reference/request/v4.examples.markdown
cbobrie: Hi Charlie. There seems to be some current discussion wth Carol on this subject, however as we are externalizing the Get list of Request this saturday, and knowing that removing the curls will require dev time not planned and possible before saturday - of confirmed, please display the documentation with Curl, and for the october release (there will be new documentation update) we'll see what is really necessary and what can be done on the dev side to make it possible.
Thanks in advance.

PR #1522, opened 2018-09-05
Update 01-profile-resource.markdown

joshuaconcur:
Added scope detail which was missing for Company ID.


joshuaconcur:
Upon onboarding American Airlines we found a discrepancy in our documentation.

PR #1532, opened 2018-09-06
UX Guidelines - adding consumer doc and updating enterprise

brittnwestSAP:
Added the consumer ux guideline doc in markdown. Updated the enterprise doc to align and updated several images.


brittnwestSAP:
Note that these still have references to docs and pdfs that may require migration as well...

PR #1567, opened 2018-09-22
Replace the screen shot: app-center-ux-guidelines-enterprise-account-creation-small.png 

dstaab:
app-center-ux-guidelines-enterprise-account-creation-small


dstaab:
deferring to BWest for this update since I do not have the access to upload the replacement image.

PR #1570, opened 2018-09-22
Delete app-center-ux-guidelines-enterprise-account-creation-small.png

dstaab:


dstaab:
deleting original screen shot that did not have checkboxes for T&Cs and Privacy Policy.  I will upload a new image with the same name as the old.

dstaab:
deferring to BWest for this update since I do not have the access to upload the replacement image.

PR #1590, opened 2018-09-25
Deprecate old hs docs

wojtekkedzior:
Remove initial version of Hotel Service 2 docs and mark all Hotel Service 1 docs as deprecated


charlieconcur:
@wojtekkedzior This change is going to be_really_ hard to review and manage. For some reason the small changes (like adding the `{ %deprecation_alert% }` are showing up as wholesale changes rather than specific line changes -- so it's a sea of red and green.

I'm also not sure we should outright remove version 1 docs if those are tied to a version 1 of the API; If those docs are outdated and we have new (better) docs which replace them, that's cool.

So, let's talk this through and see how we can move forward.

Chances are we will want at least two pull requests to cover this change:

1) Remove files.
2) Edit files.

wojtekkedzior:
Good points Charlie.  I will scrap this PR and come up with two separate ones for removing and editing of the file.  I'll double check why the one line change registers as a full file change.  Thinking it might be the tool i'm using.

PR #1601, opened 2018-09-28
Prerelease blurb

charlieconcur:
This change creates an includes file containing the pre-release text and updates all instances of 'preview' or 'prerelease' or 'beta' to use this includes HTML.


src/_includes/prerelease.html
```diff
@@ -0,0 +1,3 @@
+<div>
+  <p><strong>Warning</strong>: This API is currently in pre-release status and is only available to approved early access partners. The API is under development and might change before being generally released. To become an early access partner, contact your SAP Concur Representative.</p>
```
sarraloew: As discussed, let's change this from Warning to Important.

sarraloew (changes requested):
Just changing Warning to Important.

src/_includes/prerelease.html
charlieconcur: Changed.

PR #1630, opened 2018-10-11
Updated Oct RNs and posted announcement

sarraloew:


src/announcements/index.md
```diff
@@ -3,6 +3,10 @@ title: SAP Concur Developer Center - Announcements
 layout: reference
 ---
 
+2018-10-11 :: **Invoice API Release Notes Published**
```
charlieconcur: Are we sure we want to make it a habit of announcing release notes? Seems like it will drive down the value of these announcements due to noise.

PR #1632, opened 2018-10-12
Remove Partner details from sample requests

KetanShrikhande:


charlieconcur (changes requested):
Small little fix needed.

src/api-reference/locate/userLocations.markdown
```diff
@@ -248,7 +248,7 @@ Authorization: Bearer {token}
           "locationId": 2,
           "locationName": "Seville Spain",
           "timezoneId": "",
-          "startDate": "2019-08-23T12:00:00",
+          "startDtDate": "2019-08-23T12:00:00",
```
charlieconcur: This doesn't appear to be the right key name, according to the schema.

charlieconcur (changes requested):
Need to fix the key name issue.

PR #1635, opened 2018-10-15
Updated Purchase Req with front matter

sarraloew:


src/api-reference/invoice/purchase-request-endpoints.markdown
```diff
@@ -1,21 +1,44 @@
...
-title: Purchase Request
+title: Purchase Request v4
 layout: reference
 ---
 
 {% include prerelease.html %}
 
-# Purchase Request
+The Purchase Request API gives clients the ability to leverage external data to create Concur Invoice purchase requests for pre-authorization. Clients and Partners can build a direct connection where data can be fed to the purchase request API which will create new purchase requests in Concur. These API-created purchase requests are automatically submitted into the Concur Invoice pre-authorization workflow. Once approved, it will result in a Concur Invoice purchase order that can be transmitted to your Vendor from Concur Invoice. 
+In addition a Get Status endpoint is provided that can be used to get basic information and check the status of the created purchase request.
```
charlieconcur, line 10: Need a line break here -- otherwise it's run on.

src/api-reference/invoice/purchase-request-endpoints.markdown
```diff
@@ -1,21 +1,44 @@
...
+## <a name="scope-usage"></a>Scope Usage
+
+Name|Description|Endpoint
+---|---|---
+`PurchaseRequest.Read`|Read only access to purchase request data|Get details of a purchase request
+`PurchaseRequest.Write`|Read and Write access to purchase request data|Create a new purchase request
+
+## <a name="dependencies"></a>Dependencies
+
+No other SAP Concur APIs are required to use this API.
```
charlieconcur: Might I suggest simply saying 'None' here is more elegant than an entire sentence? Also, dependencies can take forms other than APIs, such as code libraries.

```suggestion
None
```

src/api-reference/invoice/purchase-request-endpoints.markdown
```diff
@@ -1,21 +1,44 @@
...
+`PurchaseRequest.Write`|Read and Write access to purchase request data|Create a new purchase request
+
+## <a name="dependencies"></a>Dependencies
+
+No other SAP Concur APIs are required to use this API. 
+
+## <a name="access-token-usage"></a>Access Token Usage
 
-4.0
+This API will work with both Company or User JWT, however Company JWT is recommended for integrations using this API if the end goal is for the integration to create purchase requests for multiple requestors. When using user JWT to create purchase requests through the API it will result in the purchase request being assigned to the user that generated the user JWT and it will not honor the user set in the payload. User JWT could be used for testing to check if your payloads are good if needed. For an integration as mentioned above company JWT is the best choice.
```
charlieconcur: We really need to get away from using the term (acronym) `JWT` when what we really mean is `access token`.

charlieconcur (changes requested):
See comments in line.

src/api-reference/invoice/purchase-request-endpoints.markdown
sarraloew: It looks like that term is all over the portal. Can we do a transitional thing and use "access token (also known as JWT)"?

charlieconcur: Not really because it's not an accurate statement.

And, yes, many teams have this very confused.

Access tokens can take many forms and one of those forms are a JSON Web Token (JWT).
Some access tokens are JWTs.
Some JWTs are access tokens.
So, if someone gives you a JWT it doesn't mean it is an access token.
A JWT can be any data whatsoever.

At some point in the future we could change the definition of access token (we already did this once) and move away from JWT.

charlieconcur (approved):

PR #1649, opened 2018-10-25
TVR-9203 Request API v4 documentation update

laurentlhermine:


charlieconcur (changes requested):
These changes are really good ones! We need to fix the whitespace issues in the JSON examples -- replacing tabs with two spaces. Otherwise the samples bleed off the page more than they should.

cbobrie:
Hi @laurentlhermine could you please check the tab space in the example file? I've checked them and they did not have any tab whitespaces but only two spaces, that's weird.
Appart from that it seems that we are ok with the rest of the changes 👍

charlieconcur:
You can see it really clearly in this JSON: https://github.com/concur/developer.concur.com/blob/4c0990eef12fe1d191db00d621b8d21a2554446a/src/api-reference/request/v4.examples.markdown#http-response-1

charlieconcur (approved):

PR #1663, opened 2018-11-02
CLQ-73708 HS2 - HSAPI V1 - Trim down unused nodes

varvaragabbarelli:


charlieconcur:
@FrollsV Let us know when you are finished with your commits so we can merge. I don't want to do it before you are ready. ;-)

varvaragabbarelli:
@charlieconcur  sorry for those commits :) We were doing review with my colleagues. You can merge it now

PR #1710, opened 2018-11-22
Removing eid from response

UdayKumarNettem:


sarraloew (approved):

sarraloew:
:shipit:

src/api-reference/expense/travelallowance/travelallowance-days.md
```diff
@@ -122,7 +117,8 @@ Name|Type|Format|Description
 ```shell
 GET https://us.api.concursolutions.com/api/v4/
 travelallowance/allowancedays/companyUUID/{companyUUID}/contexts/{Context}/{ContextId}
-Accept: application/json
+Authorization: Bearer {token}
+Content-Type: application/json
```
charlieconcur, line 121: It's unusual to see a payload with a GET request. See https://tools.ietf.org/html/rfc7231#section-4.3.1. I will also note this payload appears to duplicate the items in the URI, making it redundant. Are we sure this is how the API works? If so, not a great design decision.

charlieconcur (changes requested):

src/api-reference/expense/travelallowance/travelallowance-days.md
UdayKumarNettem: We need to pass access_token and type as a part of company JWT validation. This API works with company JWT

PR #1713, opened 2018-12-01
Fixed Tip formatting

RoAnnC1:
@howarddierking The numbered list is supposed to be 1-6, but it renumbered when it was pushed. I've tried to fix it, but am not sure if I did. All the various previews I looked at, including Atom show it rendering properly. Can you advise if it doesn't work?


sarraloew (approved):

sarraloew:
:shipit:

Issue #36, opened 2015-03-25
Links go to Drupal site expense/attendee/attendee-list-resource.html page


henryyamamoto-concur:
http://concur.github.io/developer.concur.com/api-reference/expense/attendee/attendee-list-resource.html
- All links on this page go to Drupal site

PR #1731, opened 2018-12-11
Adding links and currencyCode to API Response

UdayKumarNettem:


sarraloew (changes requested):
Please add definitions for the rel and href fields that you've added to the example. 
Also, please remove the extra line at line #36, it's breaking the table.

PR #1748, opened 2019-01-08
Update intro language for Partner Support page.

RoAnnC1:


src/tools-support/requesting-partner-support.markdown
```diff
@@ -3,12 +3,15 @@ title: Request Support - Partners
...
 
-[Login and open a partner support case](https://sapconcur.my.salesforce.com/secur/login_portal.jsp?orgId=00D600000007Dq3&portalId=06060000000PrEi)
+## How To Log a Support Case
 
-**App Center Partners should never instruct a customer to log a ticket directly into SAP Concur Customer Support**
+In the App Center Partner Program, partners provide first-line support for their application. 
 
-* The Partner must log the case if the issue is related to the integration between SAP Concur and the Partner.
-* If you are an established App Center Partner and do not have log in credentials to the partner support portal please send an email to  [App Center Alliance](mailto:concur_appcenteralliance@sap.com?Subject=Support%20Portal%20Credentials) to request.
+* Please do not advise a customer to log a ticket directly with SAP Concur.
```
charlieconcur: That's probably not the customer-friendly message we want to send?

RoAnnC1: Well, it doesn't seem any less friendly than what was at the top of the page originally:
*App Center Partners should never instruct a customer to log a ticket directly into SAP Concur Customer Support**

At least we say please...

This is the language Jessica requested.

charlieconcur: I like what you've said about partners providing first-line support. That's super elegant. I'm balking at the second half (which, yes, I see in the original). If one of our customers is having a problem i don't think we want to send a negative message about contacting support.

Something like this would be better: "When customers open a support ticket with SAP Concur for your app it will unnecessarily slow the support process. We are happy to help them as our shared customer -- and it will likely drive their satisfaction down."

...or some such friendlier way of saying things.

RoAnnC1: I showed your proposed language to Jessica since I don’t really own the messaging. She was fine with tweaking, but thought this might be too long. 

How about “Please do not advise a customer to log a ticket directly with SAP Concur, doing so may slow the support process down and reduce customer satisfaction.”

I think that gets the point across about why not to do that rather than just say NO. And it’s a little more compact.

charlieconcur: How about: “Please advise customers to contact you directly for support rather than SAP Concur which may slow the support process down and reduce customer satisfaction.”

charlieconcur:
I see we are adding new files -- make sure they are versioned. ;-)

src/api-reference/financial-integration/financial-documents-schema.markdown
```diff
@@ -0,0 +1,301 @@
...
+* [Cash Advance](#sccashadv)
+  * [Cash Advance - employeeData](#sccaemp)
+  * [Cash Advance - cashAdvanceData](#sccaadv)
+  * [Cash Advance - journalData](#sccajour)
+
+## <a name="#scexpense"></a>Expense
+
+### <a name="#scexpemploy"></a>Expense - Employee
+
+| **Name**                                            | **Type**    | **Format** | **Description**                                                                                                                                            |
```
charlieconcur: We will need to simplify this table formatting.

PR #1755, opened 2019-01-14
Invoice Pay API changes

mahimarpi:


sarraloew (approved):

sarraloew:
:shipit:

charlieconcur:
@mahimarpi Before I commit I would like to understand the changes to the API in removing the `companyId` and the dependency on the profile service no longer necessary. Thoughts?

charlieconcur:
I am going to go ahead and merge this pull request noting we are likely going to confuse readers because apparently we are going to bury what should go into the query string portion of the URI is now being pulled out of the access token (which has the form of a JWT). Without making any value judgement about the validity of this design we need to circle back in another pull request to document exactly how the service determines what data is returned.

**Before**

```
GET https://us.api.concursolutions.com/invoice/provider-payment/v1/payments?companyId=123456789
```
Equivalent to SELECT * FROM payments WHERE companyId=123456789

**After**

```
GET https://us.api.concursolutions.com/invoice/provider-payment/v4/payments
```

Equivalent to SELECT * FROM payments -- presumably the companyId is extracted from the data within the access token.

PR #1770, opened 2019-01-18
Fix whitespace in fin int schema

charlieconcur:
This pull request corrects the whitespace throughout the document, especially in the tables, to match the documentation template.


src/api-reference/financial-integration/v4.financial-documents-schema.markdown
```diff
@@ -28,311 +28,312 @@ layout: reference
...
+`addressName`|String|-|Address Name.
+`city`|String|-|City defined on the Ship To Address.
+`contactFirstName`|String|-|Ship To Contact First Name.
+`contactLastName`|String|-|Ship To Contact Last Name.
+`countryCode`|String|-|ISO Country Code.
+`phoneNumber`|Numeric| Integer| Phone Number on address.
+`postalCode`|String|-|Postal Code value.
+`state`|String|-|State value on Ship To Address.
+`taxJurisdictionCode`| String|-|The Tax JurisdictionCode value defined on the Ship To Address.
+`poLineNumber`|Numeric|-|The PO line item number associated to the payment request. Maximum length 10 characters.
```
charlieconcur, line 281: @RoAnnC1 Can you verify if `poLineNumber` should be here? Thanks.

src/api-reference/financial-integration/v1.financial-integration.markdown
```diff
@@ -17,7 +17,6 @@ Below are some benefits for using the Financial Integration service:
 * Opportunity to modify and retry before the object is locked in SAP Concur.
 * Ensures data sync between systems.
 
-
```
sarraloew: The extra whitespace here is to prevent markdown from consolidating these two lists into one big list, which is incorrect. Let's add this text here to separate them (after the Ensures data sync between systems, before the link to the use cases): "This API provides one or more financial documents for each data object in SAP Concur."

src/api-reference/financial-integration/v4.Financial-Integration-Use-Cases-Extract-Information.markdown
```diff
@@ -5,7 +5,7 @@ layout: reference
 
 {% include prerelease.html %}
 
-**Note:** The extract file excerpts in this document are a reference point for those developers who have used those files for prior financial integrations. This excerpt is intended to help those developers transition from an extract file to the Financial Integration Service. This allows you to identify what you have been using from the Extract File and to locate the same data in the new JSON response. If you have not used the extract file previously, the Extract file excerpts can be skipped.
```
sarraloew: Why remove the bold? Also this edit removed the space between the : and the sentence, please re-add it.

src/api-reference/financial-integration/v4.financial-documents-schema.markdown
sarraloew: The Invoice fields are still under design so we'll be getting an updated schema for the Invoice content by end of January. For now we can publish as-is, knowing we'll update it later.

sarraloew (changes requested):

src/api-reference/financial-integration/v1.financial-integration.markdown
charlieconcur: Oh, I get it -- one is a table of contents. Got it.

src/api-reference/financial-integration/v4.Financial-Integration-Use-Cases-Extract-Information.markdown
charlieconcur: I would prefer we just remove the word 'Note' altogether -- I'm unsure of the purpose.

src/api-reference/financial-integration/v1.financial-integration.markdown
charlieconcur: I'm adding a **Contents** H2 since we have a bulleted list leading into a bulleted list.

src/api-reference/financial-integration/v4.Financial-Integration-Use-Cases-Extract-Information.markdown
sarraloew: In this case, we want to warn readers that this information is only for a subset of readers. If people who aren't familiar with the extracts read this content, it will only confuse them. So, I'm open to something like Important instead of Note, it can be anything but we need it to be eye-catching.

charlieconcur: Reverted.

charlieconcur:
@sarraloew Think I've addressed most if not all of your comments -- let me know.

PR #1777, opened 2019-01-25
Update v4.invoice-pay.markdown

mahimarpi:
We don't have dependency on Invoice Pay being enabled for the APIs.


src/api-reference/invoice/v4.invoice-pay.markdown
```diff
@@ -39,11 +39,6 @@ SAP Concur partners with external payment providers for processing invoice payme
 Name|Description|Endpoint
 ---|---|---
 `invoice.providerpayment.write`|Read access to pending payments, and write access to payment status|GET,POST
-`company.read`|Required to get the SAP Concur company identifier|GET
-
-## <a name="dependencies"></a>Dependencies
```
charlieconcur: @sarraloew Do you want to (a) leave the **Dependencies** header in the document and state 'None' or do you want to go ahead and (b) remove the Dependencies header altogether?

My preference is probably 'a'.

sarraloew: We should keep the Dependencies section. I've sent a question to Mahima about the Dependencies.

sarraloew (approved):

sarraloew:
:shipit:

PR #1782, opened 2019-01-29
Update punctuation and headings

RoAnnC1:
Charlie,
This pull request includes an update to the old page that I did a few days ago. I've made the same change in the new content. Just FYI in case you're wondering why both pages appear.


sarraloew (approved):

sarraloew:
:shipit:

PR #1791, opened 2019-01-31
Changing buttons to blue

LaurelC3:
With the new white background, the white buttons need to be changed to blue to stand out.


charlieconcur:
This isn't an issue with the color of the buttons -- it is layout wonkiness caused by prior HTML changes. Fixed with #1795.

PR #1819, opened 2019-02-08
Added note about March meeting to Business Review section

RoAnnC1:
Charlie,
Can you review my HTML and make sure I did it correctly. I want the note to stand out so I made it a paragraph. Not sure if bold works with white text so I didn't add that.


sarraloew (approved):
I dropped it into Dreamweaver and it rendered correctly, so I think it's good.

sarraloew:
:shipit:

PR #1828, opened 2019-02-14
Corrected link

RoAnnC1:
Previous link was 404 because it included "tools-support/reference/" in the path.

### DIFF
--- a/src/tools-support/reference/custom-list-items.markdown
+++ b/src/tools-support/reference/custom-list-items.markdown
@@ -7,7 +7,7 @@ Custom list fields are included in many of the web services calls throughout Con
 
 ##  Value
 
-When posting a list item, the list item **code** should be sent as the value, not the list item **name**. The code is returned in the **levelxcode** element of the [Get List Items](./api-reference/common/list-item/v3.list-item.html#get) function.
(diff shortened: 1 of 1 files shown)


src/tools-support/reference/custom-list-items.markdown
```diff
@@ -7,7 +7,7 @@ Custom list fields are included in many of the web services calls throughout Con
 
 ##  Value
 
-When posting a list item, the list item **code** should be sent as the value, not the list item **name**. The code is returned in the **levelxcode** element of the [Get List Items](./api-reference/common/list-item/v3.list-item.html#get) function.
+When posting a list item, the list item **code** should be sent as the value, not the list item **name**. The code is returned in the **levelxcode** element of the [Get List Items](https://developer.concur.com/api-reference/common/list-item/v3.list-item.html#get) function.
```
charlieconcur:
```suggestion
When posting a list item, the list item **code** should be sent as the value, not the list item **name**. The code is returned in the **levelxcode** element of the [Get List Items](/api-reference/common/list-item/v3.list-item.html#get) function.
```

sarraloew:
:shipit:

PR #1834, opened 2019-02-18
Adding PDFs

RoAnnC1:
Charlie,
We need URLs to put in a document that partners will be using. These docs are currently only available once you log in to Concur and need to be visible before that. They will not be in the TOC.


charlieconcur:
Things to resolve...

No spaces in file name.
File name in all lower case.
Need to be in something other than the root path. Based on the file name / purposes perhaps under /manage-apps.

I'm going to assume these have to stay in PDF format?

charlieconcur:
I found this already in place with a newer revision date...

https://github.com/concur/developer.concur.com/blob/preview/src/manage-apps/go-market-docs/app-center-letter-of-authorization.pdf

...which also begs the question if these should be in the `/manage-apps/go-market-docs/` path.

RoAnnC1:
Didn't realize the LOA was the link auth doc. I'll remove it.

sarraloew (approved):

sarraloew:
Looks good. We will be referencing this file on an ongoing basis, not just when partners go to market, so I think this location is a good fit for it. The PDFs are required so that no one can change the legal text. 
:shipit:

PR #1853, opened 2019-02-26
Update links to correct version of auth

RoAnnC1:
Note that the links to auth content in these files pointed to new auth, which the travel profile team has confirmed Notification services doesn't support.


sarraloew:
:shipit:

PR #1854, opened 2019-02-26
Update image paths with hyphen instead of underscore.

RoAnnC1:
These images weren't displaying due to incorrect paths.


sarraloew:
:shipit:

PR #1884, opened 2019-03-22
Move Parameters section above Response Class section

truclyle:
Moved the Parameters section above the Response Class section

@sarraloew: The Response Messages section appears to be already at the end of the list but since I cannot make any HTTP requests in the development environment, I cannot confirm 100%.


sarraloew:
:shipit:

charlieconcur:
This one will need local machine before pushing to `preview`. I'll get some instructions for you.

charlieconcur (changes requested):
We should not merge this until it has been evaluated on a local machine to achieve the desired outcome.

truclyle:
@charlieconcur: I have verified on a local machine that the changes I made produce the desired outcome, which is to display the Response Class section after the Parameters section.
 
layout

PR #1899, opened 2019-03-31
Code of conduct addition

charlieconcur:
We will need to edit the email address as appropriate. Hopefully @sarraloew already has an externally available tech pubs DL we can use.


sarraloew:
This looks like it applies to the forum- who is responsible for moderating that?

charlieconcur:
It applies to both, actually. I think Doug owns the forum.

charlieconcur:
@sarraloew Do you want to move this pull request forward?

charlieconcur:
Closing as there doesn't appear to be any support for this -- unfortunate, but a reality.

PR #1915, opened 2019-04-03
Sample code from integration team

charlieconcur:


src/tools-support/samples/index.markdown
```diff
@@ -0,0 +1,26 @@
+---
+title: Platform Code Samples
+layout: reference
+---
+
+These sample projects demonstrate how to accomplish atomic tasks as you develop your own applications and may be helpful to anyone who wants to learn to more about how to use the SAP Concur platform.
+
+The following links are to ZIP archive downloads:
```
RoAnnC1, line 8: I wasn't going to number the downloads because it implies that they need to be worked through in order, and they don't.

charlieconcur: Yes, I considered NOT numbering them and wanted to also respect how the integration folks had these on the site -- so trying to find the balance.

RoAnnC1: I asked Pedro about this and he didn't think they needed to be numbered.

sarraloew:
:shipit:

PR #1923, opened 2019-04-04
A few suggested edits to the documentation template.

charlieconcur:


charlieconcur:
:shipit:

sarraloew:
@charlieconcur These changes look fine, but I thought we were using the doc template in the other repo- https://github.concur.com/developer/api/blob/main/documentation-template.md as the standard. Are you wanting to move the doc template here?

charlieconcur:
It is in both places and I keep them synced.

PR #1936, opened 2019-04-12
PIE-2028/Removed currency field from payment object

shayanraouf:
Removed currency field from payment object in IPM payment response.


charlieconcur:
:shipit:

PR #1940, opened 2019-04-15
Added question to explain app center filtering

brittnwestSAP:
"What should I do if I cannot find an application" - per customer feedback


charlieconcur:
:shipit:

PR #1941, opened 2019-04-15
Updated regional availability filtering

LaurelC3:
Added more details to the regions section to make the filtering clearer to clients.


charlieconcur:
:shipit:

PR #1944, opened 2019-04-16
TVR-9964 Update Request documentation. New policy endpoint. Employee schema. List item schema.

laurentlhermine:

### DIFF
--- a/src/api-reference/request/v4.endpoints.markdown
+++ b/src/api-reference/request/v4.endpoints.markdown
@@ -22,28 +22,32 @@ If a new or existing project please use version [3.0](./v3.request.html) of this
   * [Get an existing expected expense](#get-existing-expense-expense-resource)
   * [Update the content of an existing expected expense](#update-content-existing-expected-expense-resource)
   * [Delete an expected expense from the Request](#delete-expected-expense-request-resource)
+* [Request Policy Resources](#request-policy-resource)
+  * [Get the list of existing Request policies for a given user](#get-user-request-policy-resource)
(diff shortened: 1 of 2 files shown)


charlieconcur:
:shipit:

src/api-reference/request/v4.endpoints.markdown
```diff
@@ -660,42 +715,21 @@ id|string|-|**Required** Unique identifier of the Expense type
...
+code|string|-|The short code of the list item
+value|string|-|Name of the Expense type
+href|string|[RFC 3986](https://tools.ietf.org/html/rfc3986)|Hyperlink to the resource for the list item
 
 ### <a name="schema-segmentleg"></a>Segment Leg
 
 Name|Type|Format|Description
 ---|---|---|---
-comment|string|-|Contains the last comment saved in this segment leg.
+class|list|[List Item Field](#schema-listitemfield)|The booking class of the segment leg, class part of the CLASS_LI_KEY table.
```
cbobrie, line 730: @laurentlhermine 
"The booking class of the segment leg, class part of the CLASS_LI_KEY table" could be eventually replaced by "The booking class of the segment leg." without necessarily adding the technical information of the table reference. 
This might brought some question if the client uses the field "class" in the segment leg linked to a custom table (and not the CLASS_LI_KEY table) as it will not work, but even in the audit rule technical guide there is the class field and class of service field respectively referring to the class of the segment leg and class of service of the segment leg. Let's delete the CLASS_LI_KEY reference in the comment and we'll see if this raises some use cases.

PR #1945, opened 2019-04-16
Update why-concur.html

cskelley:
Added explanatory text to better qualify/filter requests and to set expectations of prospective ISVs.


charlieconcur:
:shipit:

PR #1961, opened 2019-04-18
Updated text to "SAP Concur"

LaurelC3:
Updated requirements to say "SAP Concur" to avoid partners using the wrong name on their landing page.


charlieconcur:
:shipit:

PR #1966, opened 2019-04-22
Sign in with Concur for Enterprise Applications

brittnwestSAP:
Describing enterprise application integrations in more detail. Providing guidelines for email verification for companies using email address as a unique identifier (not unique in Concur so requires verification before first sign in).


charlieconcur (changes requested):
Do not use underscores (`_`) in file names -- use a dash instead (`-`).

brittnwestSAP:
updated new files and migrated old to dashes rather than underscore. Also changed full path to use dash - this is currently in pre-release and only affects 2 partners (whom I will notify specifically of the path change). Added pre-release header as well.

charlieconcur (approved):

charlieconcur:
:shipit:

PR #1979, opened 2019-05-06
 Modify to accept employee ID in lieu of email for an employee

elankumaranrk:
Budget endpoints should accept employee ID in place of email ID/employee CUUID to identify an employee.


sarraloew:
:shipit:

PR #1984, opened 2019-05-09
Update Vendor Group Endpoint URI in Document

HNozakiC:
Functionally, Endpoint URI for PUT and DELETE method is /api/v3.0/invoice/vendors/groups
Please update the document. Thank you!


HNozakiC:
Sorry, I need more test before I request this.

PR #2008, opened 2019-06-12
Event Subscription new description

usandre:
The old one in event-pubsub/ folder will be deleted after this revision and submitted in separate PR


usandre:
@charlieconcur can you please review this?

PR #2015, opened 2019-06-20
Updating activation guide to clarify navigation

brittnwestSAP:
The prior guide left out important steps to find the app and what the app looked liked when an administrator can connect. Adding in those steps.


jonikamelcher:
:shipit:

PR #2072, opened 2019-09-05
EXP-29565: Remove the Cost Object query string parameter from the API documentation

rajatkasliwal:
* Removed cost object query string from the API reference section of the developer.concur.com site

* Removed cost object query string from the API explorer section of the developer.concur.com site


jonikamelcher:
:shipit:

PR #2078, opened 2019-09-09
Updating Logo Usage Guidelines

LaurelC3:
Updated logo usage section for clarity as the Approved checkmark did not accurately represent that the logos may only be used in specific cases and with prior approval.


RoAnnC1:
:shipit:

PR #2085, opened 2019-09-11
Updated Logo Guidelines Section Formatting

LaurelC3:
Removed the checkmarks and some text for a cleaner look.


RoAnnC1 (changes requested):
Laurel,
Now that there's no checkmark can we update so the text isn't centered in the first column? It looks odd and is harder to read that way.

LaurelC3:
Yes – absolutely. Feel free to re-format however looks cleanest!

RoAnnC1:
I don't have permission to change your pull request. All you need to do is remove the colon from this line. That's what makes the text centered.
:---:|---

Instead it should be
---|---

LaurelC3:
I'll create a new pull request with correct formatting to remove centered text.

PR #2086, opened 2019-09-12
Update PolicyID Description

HNozakiC:
User's default expense policy is being used when creating a new expense report


sarraloew:
:shipit:

PR #2108, opened 2019-09-30
Update v3.vendor.markdown

shinyay-jptc:
IsLineItemVatIncld is not listed in Vendor Schema of API Reference Page. In the swagger, we can see it.


sarraloew:
:shipit:

PR #2125, opened 2019-10-14
ESS walkthrough first version

usandre:


usandre:
@haydenlam there are couple question marks, can you please help me with those?

jonikamelcher:
Hi @usandre, for reference, the documentation template can be found here: https://github.concur.com/developer/documentation/blob/master/documentation-template.md. I understand this is the first draft but just wanted to pass it along now because we will need all the information in the template in order to do the final review. I'll go ahead and get these set of changes merged shortly.

jonikamelcher:
:shipit:

PR #2128, opened 2019-10-14
Clarify www- URI for Authorization Grant

haydenlam:


RoAnnC1 (changes requested):
On line 289, "inorder" should be two words.

haydenlam:
Corrected.

RoAnnC1:
:shipit:

PR #2137, opened 2019-10-17
APIs for publishing on Internal SAP API Hub: List-Service

melissa-von-wald:
Swagger JSON files for list service APIs
https://jira.concur.com/browse/EXP-29964


sarraloew:
:shipit:

PR #2145, opened 2019-10-24
EXP-29963 - first draft of swagger for report forms api

jtbigwoo:
EXP-29963 - first draft of swagger for connect api report forms endpoints


sarraloew:
:shipit:

PR #2165, opened 2019-11-07
Add and remove Required to/from several fields

HNozakiC:
To create a new expense entry, ReportID, ExpenseTypeCode, TransactionDate, PaymentTypeID, TransactionCurrencyCode and TransactionAmount are the required field.  ExpenseTypeName, ReportOwnerID, SpendCategoryCode and SpendCategoryName are not necessarily to create a new expense entry.
Also, 'user' is optional parameter when creating a new expense entry.  If the access token is not obtained by the user who owns the entries, 'user' parameter is required.


jtbigwoo (approved):
Looks good to me

jonikamelcher:
:shipit:

PR #2185, opened 2019-11-22
Add response 400 error codes info

arturoduar:
Additional info comes back after the API resturns a 400 response code. This PR adds what they mean.


jonikamelcher:
:shipit:

PR #2198, opened 2019-12-13
Replaced token/ids with dummy values

MarcBell-SAP:
dummy values are 000..., 111...., 222...., 333....

Also fixed up some minor labelling errors. Client_ID was being used instead of client_secret in a number of locations.


RoAnnC1:
:shipit:

PR #2213, opened 2019-12-25
expenseTypeId is required in payload.

shinyay-jptc:
When I post Quick Expense v4 data, I got following message. It looks ExpenseTypeId is required.

    "httpStatus": "400 - Bad Request",
    "errorMessage": "Request received without ExpenseTypeId",


jonikamelcher:
:shipit:

PR #2246, opened 2020-02-18
Please update supported image types

shinyay-jptc:
>`fileContent`|`file`|-|**Required** The quick expense image. Maximum size 5 MBs. Supported image types are: PNG, PDF, TIFF, JPEG

GIF file is not supported in UI. (PNG, JPG, JPEG, PDF, TIF, TIFF file for upload are valid.) So, description should be same as Receipt Upload in UI.


jonikamelcher:
:shipit:

PR #2260, opened 2020-03-05
TVR-11372 remove Request report creation from public api

quancheng-zhao:
Creation Expense report from Request is not ready to be public yet, so we remove this endpoint from public API.


jonikamelcher:
:shipit:

PR #2273, opened 2020-03-10
Reverting code fix - was an admin panel issue

MarcBell-SAP:
Reverting the code fix because the problem was actually in the admin panel in google tag manager. Google Analytics by default records page views and Google Tag Manager was mistakenly setup to record page views as well.


jonikamelcher:
:shipit:

PR #2312, opened 2020-04-29
remove the PasswordPolicy from the API response

suhita-banerjee:
https://jira.concur.com/browse/CP-2267


jonikamelcher:
:shipit:

PR #2319, opened 2020-05-01
Make updates per product owner suggestions

ledbutter:
Jira [reference](https://jira.concur.com/browse/BUD-2325)

### DIFF
--- a/src/api-reference/budget/getting-started.md
+++ b/src/api-reference/budget/getting-started.md
@@ -52,6 +52,9 @@ Name|Description|Endpoint
 `fiscalcalendar.read`|Grants read access to the fiscal calendar.|GET All Fiscal Years, GET a Fiscal Year
 `fiscalcalendar.write`|Grants read and write access to the fiscal calendar.|GET All Fiscal Years, GET a Fiscal Year, POST a Fiscal Year, DELETE a Fiscal Year
 
+## <a name="dependencies"></a>Dependencies
+
+SAP Concur clients must purchase Budget in order to use this API.
 
 ## <a name="access-token-usage"></a>Access Token Usage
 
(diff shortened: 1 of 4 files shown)


src/api-reference/budget/getting-started.md
```diff
@@ -52,16 +52,6 @@ Name|Description|Endpoint
...
-This API uses this optional scope:
-
-Name|Description|Endpoint
----|---|---
-`LIST` |GET List Items, necessary for budget tracking value `listKey` if using list fields for budget tracking. |POST Budget Item
-
-## <a name="dependencies"></a>Dependencies
-
-SAP Concur clients must purchase Budget in order to use this API. The developer may use the following SAP Concur API to get optional information:
-* List Item v3.0, to retrieve the budget tracking value `listKey`
```
jonikamelcher, line 64: @ledbutter Hello. Please keep the Dependencies section with the text: "SAP Concur clients must purchase Budget in order to use this API." Thanks.

jonikamelcher (changes requested):

src/api-reference/budget/getting-started.md
ledbutter: Done, budget is going out of pre-release, is this the place to remove the `{% include prerelease.html %}` reference from [Budget.markdown](https://github.com/concur/developer.concur.com/blob/preview/src/api-explorer/v4-0/Budget.markdown) or is that done automatically as part of the migration to being in GA?

jonikamelcher: Thank you! We will be removing the prerelease reference as part of the migration a little later.

jonikamelcher (approved):

jonikamelcher:
:shipit:

PR #2335, opened 2020-05-19
Update provisioning_swagger.json

mvahlsing98117:
Combined swagger document for client preview. This is a makeup of profile, spend and travel swagger documents for the provisioning of a user into CTE.


MarcBell-SAP:
:shipit:

PR #2372, opened 2020-07-09
Updating User Provisioning document and adding a new flow diagram

mvahlsing98117:


jonikamelcher:
Hi @mvahlsing98117. We shouldn't be removing any sections here. If there are endpoints that are not applicable, those are fine to remove but we cannot remove sections for status codes, headers, payloads, etc... These are all part of the documentation template and we need to have this information available for users. Thank you.

jonikamelcher:
:shipit:

PR #2375, opened 2020-07-13
Please add Comment field in the schema

shinyay-jptc:
In the swagger page, it describe developer can post the comment. However, there is no comment field in the schema.
https://developer.concur.com/api-explorer/v3-0/Entries.html


jonikamelcher:
:shipit:

PR #2400, opened 2020-07-27
Update v4.user-provisioning.md

mvahlsing98117:
Removal of US only data center


mvahlsing98117:
Removed US only data center request.

jonikamelcher:
:shipit:

PR #2402, opened 2020-07-28
Role clarification

sarah-williamson:
Updated/moved content around to make the role access info clearer.


jonikamelcher:
:shipit:

PR #2414, opened 2020-08-06
ERS-2640 - Remove references to expense.report.* scopes

greg-grove-sap:


jonikamelcher:
Hi @greg-grove-sap. Could you please include an example of the delete action? If there are none, please just include an N/A. It prevents users from thinking we've missed/forgotten any fields. Thank you.

greg-grove-sap:
@jonikamelcher Added N/A, since there are no longer any "delete" scopes which can be used as an example.

jonikamelcher:
:shipit:

PR #2415, opened 2020-08-07
Add the first level children items of a list endpoint

melissa-von-wald:
Request V4 API is dependent on List V4 API
The TPMs for both teams have determined that the GET /list/v4/lists/{listId}/children endpoint to return the first level children items of a list is necessary to release as EA.


melissa-von-wald:
Leaving PR in draft state to give my team a chance to review

jonikamelcher:
:shipit:

PR #2421, opened 2020-08-14
Updating Provisioning swagger document with updated information

mvahlsing98117:
Updating Provisioning swagger document with updated information. There were some data fields that were not displayed within the document.


jonikamelcher:
:shipit:

PR #2439, opened 2020-09-08
HOTEL-2454: Fix credit card code in Hotel Service 2 API Docs

stuart-ng-van:
- Replace MC with CA for master card
- Replace DN with DC for diners club
- Added examples with explicit mapping


jonikamelcher:
:shipit:

PR #2445, opened 2020-09-18
Updating branding

LaurelC3:
Updating some uses of "SAP Concur" to "SAP Concur solutions" to comply with brand guidelines.


RoAnnC1:
:shipit:

PR #2462, opened 2020-10-07
List max corrections

sarah-williamson:
Updating the limit of list items to accurately reflect the max of 100 and updating the example to align.


jonikamelcher:
:shipit:

PR #2470, opened 2020-10-12

khushbooba91:
Cellphones will only contain digits, dashes, and spaces from the input. In UI, that rule is enforced we are going to accept only one type of input to the cellphone numbers from all the platforms.


jonikamelcher:
:shipit:

PR #2474, opened 2020-10-19
LEU v4 clarifications

sarah-williamson:
Adding that mobile is supported and clarification of language code parameter.


jonikamelcher:
Hi @sarah-williamson . Is this available only with the NexGen UI?

sarah-williamson:
Hi @jonikamelcher - Yes, for web the v4 LEU is only available for NextGen UI (the v1 works with Gateway UI). I realized that we we missing that we also support mobile with the v4, so that was the intent of the correction on the launch-external-url-v4.markdown page.

jonikamelcher:
Hi @sarah-williamson. This is a new type of notation for us, so I appreciate the patience as we discuss the best way to notate this specific scenario. One thing we do have standardized that in reference to the mobile app. If you could please update this to say: The SAP Concur Mobile App, I would appreciate it. We should have this all merged through by early next week, we just need to verify with Sarra Loew, who is out this week. Thank you.

sarah-williamson:
Thanks Jonika. Per our quick chat, I'll wait to make the mobile verbiage update until I hear back on the NextGen verbiage (in case a change is needed for that).

jonikamelcher:
Hi @sarah-williamson. If you'd be willing to update the mobile information to just: "The SAP Concur mobile app", I think we're good to go! Thank you!

sarah-williamson:
Thanks @jonikamelcher. I've made the update.

jonikamelcher:
:shipit:

PR #2475, opened 2020-10-20
Swagger and Markdown file added for Cash Advance Public API.

gaganpsgithub:


gaganpsgithub:
Created one new PR https://github.com/concur/developer.concur.com/pull/2482 for the same.

PR #2481, opened 2020-10-21
Updating user provisioning documentation. 

mvahlsing98117:
Made some updates to the user provisioning documentation in preparation for customer viewing and integration with SAP projects. 
The latest swagger needs to be pulled in as well. 

Thanks

### DIFF
--- a/src/api-reference/user-provisioning/v4.user-provisioning.md
+++ b/src/api-reference/user-provisioning/v4.user-provisioning.md
@@ -13,8 +13,12 @@ The User Provisioning Service allows callers to provision a user in the SAP Conc
 * [Scope Usage](#scope-usage)
 * [Dependencies](#dependencies)
 * [Access Token Usage](#access-token-usage)
-* [Using the /Bulk endpoint for provisioning and managing identity data](#create-bulk-provision-post-request)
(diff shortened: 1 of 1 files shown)


mvahlsing98117:
Making updates to the use provisioning documentation. Want to verify that the updates fit within Tech Hub's publishing guidelines.

jonikamelcher:
Hi @mvahlsing98117. We need this documentation broken out a little further. Each PUT, POST, PATCH, is a separate action and thus needs it's own section of the documentation template. We cannot have all the actions under one portion because it tends to cause confusion. If you could please break these out into their own Create a resource/Endpoint Documentation portion of the template, we can complete the review. It would also be helpful to know in what situations each would be used (Is POST like a create? PUT an edit? PATCH...) Thank you.

mvahlsing98117:
Made some changes based on the suggestions above. Thanks for the input and let me know if more changes are needed.

jonikamelcher:
Hi @mvahlsing98117. Looking great! I will take a deeper look into this early next week and get it merged through. Thank you!

jonikamelcher:
Hi @mvahlsing98117. I'll be merging this through shortly and then adding some formatting changes in the next couple days, so if you see the page change a little bit, it is expected. Thank you!

PR #2492, opened 2020-11-10
Update v4.reports.markdown

visadin:
Added two more attributes submitterId and taxConfigId that have been introduced to the GET reportById schema based on new requirements.


jonikamelcher:
:shipit:

PR #2500, opened 2020-11-17
Update v4.reports.markdown

visadin:
Minor change to make sure that description lines up in the right column


jonikamelcher:
:shipit:

PR #2504, opened 2020-11-19
Removed includePersonalTrips from query parameters

roshan-kumar001:
`includePersonalTrips` query parameter has been removed from the endpoint.


jonikamelcher:
:shipit:

PR #2515, opened 2020-11-27
Change Base URI for Example

HNozakiC:
Change Base URI from 'www.concursolutions.com' to 'us.api.concursolutions.com'. Because 'www.concursolutions.com' sometime respond  503 Service Unavailable.


jonikamelcher:
:shipit:

PR #2530, opened 2020-12-08
LEU Updates

sarah-williamson:
Minor text updates and added username/password length requirements


sarah-williamson:
Also includes adding scopes/grants info per training feedback and clarifying configuration process to add more details about the v4 process

jonikamelcher:
:shipit:

PR #2544, opened 2020-12-28
Update index.markdown

Premysl-Stuksa:
Added a note that updating a password via update function of this API will not update password expire date on the DB. 
This results in users still having to change their password again when logging into SAP Concur after the update via API.


jonikamelcher:
:shipit:

PR #2546, opened 2021-01-05
Update List Item API

melissa-von-wald:
Update the List Item API diagram to include the delete APIs
Update the description for the isDeleted parameter


jonikamelcher:
:shipit:

PR #2553, opened 2021-01-08
Add missing image for list item

melissa-von-wald:
In my last PR I mistakenly deleted the old image and missed committing the new image


jonikamelcher:
:shipit:

PR #2588, opened 2021-01-27
Document HTTP 429 Responses

aweiker:
This can be returned from various web services and is already being
used. There will be more services in the future that will use this
repsonse so this adds that to the common respnose codes for future
knowledge/documentation.


jonikamelcher:
:shipit:

PR #2602, opened 2021-02-15
EKPZH-529 - adding a note about policy to v3 reports doc

jtbigwoo:
EKPZH-529 - adding a note to explain that the policy cannot be changed even with the PUT call after it was pointed out that the language is not clear.  Previous note only referred to the POST call, but the PUT cannot change the value either.


jonikamelcher:
:shipit:

PR #2609, opened 2021-02-18
ERS-2262 - Add compact model for PATCH expense short url

greg-grove-sap:
- Add compact model for PATCH expense short url
- Update descriptions for PATCH expense/report endpoints


jonikamelcher:
:shipit:

PR #2628, opened 2021-03-02

irina-sap:
Updated Requested Parameters section for recent changes:
* LastModified is required
* Added optional ConfigID parameter


jonikamelcher:
:shipit:

PR #2633, opened 2021-03-09
DO NOT MERGE - feat: HOTEL-3054 Update HS2 API doc for rate categories

stuart-ng-van:


src/api-reference/direct-connects/hotel-service-2/Availability.markdown
```diff
@@ -79,6 +80,10 @@ Message to retrieved the availability of hotels.
           <HotelSearchCriteria>
             <Criterion>
               <HotelRef ChainCode="ZZ" HotelCode="111222"></HotelRef>
```
stuart-ng-van, line 82: Right now, we send one hotelRef per Criterion. Doing this will mean we need to repeat ratePlanCandidate within every single Criterion.

From schema, it does seem we should be able to send multiple HotelRef within a single criterion object. This will be better as no repeat of RatePlanCandidate. But not sure if current vendor  supports it

stuart-ng-van:
Some thoughts

- Do we add table and limit what rate plan type we support?

I am thinking just match with we have in UAPI for now

4 - corporate
8 - government
9 - military
14 - senior

There is no clear AAA mapping, we can use 1 - association and text AAA?

this will need some chatter with vendors.

src/api-reference/direct-connects/hotel-service-2/Availability.markdown
stuart-ng-van: I will do a test locally and see

stuart-ng-van: tested, the big four vendors only expect 1 hotelRef inside each Criteria. So either vendors need to change, or we pass in RatePlanType inside each Criterion

src/api-reference/direct-connects/hotel-service-2/Availability.markdown
```diff
@@ -187,7 +199,7 @@ The maximum allowed size of `OTA_HotelAvailRS` is 5 MB. Any response that exceed
             </RoomType>
           </RoomTypes>
           <RatePlans>
-            <RatePlan RatePlanID="XNFYP4I" AvailabilityStatus="ChangeDuringStay">
+            <RatePlan RatePlanID="XNFYP4I" AvailabilityStatus="ChangeDuringStay" RatePlanType="4">
```
stuart-ng-van, line 202: CWT is already using RatePlanType now, but not compliant to OTA standards

RatePlanType="Refundable" which is not correct.

PR #2640, opened 2021-03-10
Update v4.user-provisioning.md

irina-sap:
Added/updated information for the Travel Extension, including adding a Traver user GET section, updated Travel User Schema, added Travel scopes.


jonikamelcher:
Hi @irina-sap. Alyson has asked this live in its own doc. I will be extracting the information from this PR and creating a separate doc to merge so you may see this one close. I will provide you with the preview link to the doc in Slack when it is ready.

PR #2648, opened 2021-03-16
Added v4 Expense scopes

visadin:
The v4 Expense scopes are required to access the v4 Reports, Expenses and Allocations API.


jonikamelcher:
:shipit:

PR #2703, opened 2021-04-08

irina-sap:
Added a list of canonical values for Preferred Language field, along with a note in that field in the table. If you have a suggestion on a better way to format that list, I'm open to suggestions. Note that this is the SAME list that Suhita said she added to the Profilev4 API  - both APIs are enforcing the same set of values. It might make sense to put them on their own page and link both API docs to that page. That way there's only 1 source to maintain. But that might be a bigger project than it's worth given all the other priorities.


jonikamelcher:
:shipit:

PR #2728, opened 2021-04-21
USERPROV-364 Update document for Spend role code Definition

eddyclhung:
## What does this PR do?

Updates the Spend role code definition for the following requests accordingly.
```
1. Should we even be mentioning Internal-only fields in the Spend Role Definitions?
I have removed the descriptions because we shouldn't be offering this information
if it is not meant to be used by partners, but should these fields be removed entirely?

Response:
- We took a look at those roles, deleted 1 role, added the detail back in for 4 roles as
they are assignable via the 400 file. And leave Role Builder as is.

2. One of the Spend Role Definitions mentions Concur Insight.
This isn't an actual product name and the last usage of it in docs was 2014.
Could someone shed some light on what this actually is?

Response:
- Replaced the "Concur Insight" with "Analysis/Intelligence",
which is the same as the latest "Shr_UG_User_Admin.doc".
```

## Checklist:

 
- [ ] Public documentation change required.
- [ ] Wiki or readme.md has been changed.
- [ ] New automation tests have been created.


willmarks:
1. I think that if it is internal information only then it doesn't need to be in these docs at all? Unless would they need to know what to ask for from implementation? If there is no reason for them to know about it then I don't think they should see it as it will just be confusing. Though that might be something that Katie or Matt could answer more easily?
2. I have no idea what that is lol...

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
```diff
@@ -152,12 +151,12 @@ The role code value that can be used for provisioning with the `Spend Role Exten
...
-|`REPORTING_CONFIG_ADMIN`|Consolidation Configuration Administrator     |This role is used by companies using the Reporting Database that wish to consolidate the data from the current version and previous versions.<br>The user assigned this role can map expense types and custom fields to a single consolidated definition, and adjust other consolidation settings. <br>**NOTE:** This role does not include access to Concur Insight.|N|N|
+|`REPORTING_BUSINESS_AUTHOR`|Cognos Business Author|The user assigned this role can use Analysis/Intelligence to view data for reports submitted at all hierarchical levels within their assigned groups (as well as any data not group-related). The user is assigned the Business license type, which restricts the features to which they have access. They may also run existing reports, create new reports or modify existing basic reports using the basic tool, Query Studio. <br>**NOTE**: A user cannot be assigned this role and the Cognos Professional Author or Cognos Consumer roles, only one of these three may be assigned.|Y|N|
```
willmarks: Do we need to define terms like "Analysis/Intelligence" in terms of what they mean within concur? or are people kinda expected to know what these things are prior to coming here and already somewhat know what they are looking for. I for one don't know anything about these two things...

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
```diff
@@ -124,7 +123,7 @@ The role code value that can be used for provisioning with the `Spend Role Exten
...
 |`INV_APPROVER`|Invoice Approver|The user assigned this role can to approve invoices within an assigned group.|N|N|
-|`INV_CONFIG_ADMIN`|Invoice(Payment) Configuration Administrator|**This role is intended to be assigned to and used by SAP Concur internal staff only.**|Y|N|
+|`INV_CONFIG_ADMIN`|Invoice(Payment) Configuration Administrator|**__This role is intended to be assigned to and used by SAP Concur internal staff only, with few exceptions.__**<br>The user assigned this role can **__fully manage__** (add, edit, delete):<br><ul><li>Invoice group configurations</li><li>Policies</li><li>Invoice-based forms, fields, and validations</li><li>Invoice and authorized approver workflows</li><li>Audit rules</li><li>Expense types</li><li>Account codes and account code hierarchy</li><li>Exceptions</li><li>Image handling, including scan configurations, invoice imaging, vendor imaging</li><li>Email reminders</li><li>Company Info</li><li>Configuration change log (view only)</li></ul>**NOTE:** A user cannot be assigned the admin role and the restricted admin role.|Y|N|
```
ctruzzi: Does it make sense to also use `<li>` for `INV_CONFIG_ADMIN_RO` on their capabilities as well?

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
ctruzzi: For same area, it seems sometimes you bold the roles that cannot be also be assigned. Should `Cognos Professional Author` and `Cognos Consumer roles` be bolded?

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
```diff
@@ -152,12 +151,12 @@ The role code value that can be used for provisioning with the `Spend Role Exten
...
+|`REPORTING_PRO_AUTHOR`|Cognos Professional Author|The user assigned this role can use Analysis/Intelligence to view data for reports submitted at all hierarchical levels (no group assignment required). The user is assigned the Professional license type, which allows access to all features. They may also run existing reports, create new reports or modify existing basic reports using the basic tool, Query Studio, and create new reports or modify existing basic reports using the advanced tool, Report Studio. They can also schedule automatic report runs.|Y|N|
+|`REPORTING_CONFIG_ADMIN`|Consolidation Configuration Administrator|This role is used by companies using the Reporting Database that wish to consolidate the data from the current version and previous versions.<br>The user assigned this role can map expense types and custom fields to a single consolidated definition, and adjust other consolidation settings. <br>**NOTE:** This role does not include access to Analysis/Intelligence.|N|N|
```
ctruzzi, line 159: Is `Analysis/Intelligence` understood well enough as a product or feature we provide?

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
eddyclhung: This is a very good question, we added those format originally, but it seems like they all got rid of.

ctruzzi: Thanks, that makes more sense.

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
```diff
@@ -77,14 +77,13 @@ The role code value that can be used for provisioning with the `Spend Role Exten
...
-|`SHD_CONFIG_ADMIN`|Shared Configuration Administrator|**This role is intended to be assigned to and used by SAP Concur internal staff only.**|Y|Y|
+|`SHD_CONFIG_ADMIN`|Shared Configuration Administrator|**__This role is intended to be assigned to and used by SAP Concur internal staff only, with few exceptions.__**<br>The user assigned this role can **__fully manage__** (add, edit, delete):<br><ul><li>Feature hierarchies for Expense Reports and Vendor Invoices</li><li>Employee group configurations, including assignment of the employee form</li><li>Employee forms and fields, validations, and custom and connected lists</li><li>Expense and payment delegate configurations</li><li>Reimbursement currencies</li><li>Imaging settings</li><li>Accounting structure (ledgers)</li><li>Localization tasks</li><li>Configuration change log (view only)</li></ul>Hierarchy segment values are used to identify the hierarchy node for which this employee has approval rights. If blank, resolves to the global group.<br>**NOTE:** A user cannot be assigned the admin role and the restricted admin role.|Y|Y|
```
willmarks: When you say "A user cannot be assigned the admin role and the restricted admin role" should we explicitly say the `SHD_CONFIG_ADMIN_RO` so they know know what the restricted admin role is that is being mentioned? Maybe they are close enough together that their connection is obvious enough...

eddyclhung: I can do "A user cannot be assigned the SHD_CONFIG_ADMIN and the SHD_CONFIG_ADMIN_RO".

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
jonikamelcher: Heads up, we cannot use the html formatting. It will break the tables when it generate to the site, that was why it was removed.

src/api-reference/user-provisioning/spend/spend-role-code-definition.md
eddyclhung: As discussed in the internal team channel.
https://concur-blue.slack.com/archives/GM19MR4LF/p1619071305052700?thread_ts=1619049908.049100&cid=GM19MR4LF

jonikamelcher:
:shipit:

PR #2778, opened 2021-06-25
GEO Location and refresh token

ghost:
Added a comment to store GEO location and refresh tokens every time and use new ones.


RoAnnC1:
:shipit:

PR #2787, opened 2021-07-01
Updates to the HotelService v2 walkthrough. Fixing links to the FAQ.

rlee-sap:
Setting yaml header in xml files with layout none as loading the XML files currently leads to `Document is empty` rendering errors.


jonikamelcher:
:shipit:

PR #2790, opened 2021-07-08
TRIPSTORE-3119 - New plural route for geting trip by id endpoint "/trips/:id"

mdanyao:
Update endpoint of getting trip by id to `trips/:id` in public event and swagger doc


mantralavinay:
@mdanyao - LGTM

jonikamelcher:
:shipit:

PR #2836, opened 2021-08-19
Updating Partner Orientation Text

LaurelC3:
Updating text about the partner orientation meeting with new cadence and link.


RoAnnC1:
:shipit:

PR #2848, opened 2021-08-27
Update apidoc.markdown

cliffstrom:
Fix the format of Refresh tokens (which is just a string, not a UUID).
Also add an example of the new format.


RoAnnC1:
:shipit:

PR #2851, opened 2021-08-30
Clarifying WSAdmin role instructions

LaurelC3:
Clarifying that the WSAdmin role should be added to both Travel and Expense, per guidance from the Technical Enablement team.


jonikamelcher:
:shipit:

PR #2853, opened 2021-08-30
Revert "Clarifying WSAdmin role instructions"

LaurelC3:
Reverts concur/developer.concur.com#2851. Canceling PR due to clarification from the TE team.


jonikamelcher:
:shipit:

PR #2871, opened 2021-09-03

irina-sap:
Made Country Code optional for Cell Phone Telephone types.


jonikamelcher:
:shipit:

PR #2902, opened 2022-01-06
Updating v4.user-provisioning.md

mvahlsing98117:


RoAnnC1:
@mvahlsing98117 This is the old repo and these cannot be merged. You'll need to move them to the new https://github.com/SAP-docs/preview.developer.concur.com repo.

PR #2903, opened 2022-01-19
Update apidoc.markdown

cliffstrom:
Clarify that this section is only an example, and applications should use the base URI they receive, not always send to the US DC.


jonikamelcher:
Hi @cliffstrom. This repo has been retired. Please use: https://github.com/SAP-docs/preview.developer.concur.com. Thanks.

